PR #9, opened 2025-03-21
Refactor blog/news/feature pagination and item pages to use slugs and shared SSR-friendly component

zixben:
**This PR introduces the following key improvements:**
- Created PaginatedPublications shared component to handle pagination for Blogs, News, and Feature Articles using a common layout.
- Retained additional rendering approaches for the /blogs category to support flexibility during development:

    - page_Client-side.tsx: Client-side rendering using useState and useEffect
    - page_SSR_NO-helper.tsx: Server-side rendering with inline data fetching
    - page_SSR_WITH-helper.tsx: Server-side rendering with a dedicated helper function

      These implementations are specific to the Blogs category, but the logic can be adapted easily for News or Feature if needed. They're intended to allow developers to choose or compare different rendering patterns based on performance, SSR, or interactivity needs.
- Switched to using slug instead of id for route and API handling:
        - Updated dynamic routes (e.g., /blogs/[slug]/page.tsx).
        - Updated API fetching logic to use slug-based lookups.
- Replaced the use of .preview with .content when rendering individual publication pages, so the full article content is now displayed instead of a summary.


.gitignore
lanterno, line 1: You should always keep the .gitignore tracked in git 
check this for more information. https://stackoverflow.com/questions/5765645/should-you-commit-gitignore-into-the-git-repos

next.config.mjs
lanterno, line 1: This file and package.json should always tracked. I'm not sure why you're removing them. Without these files, a new developer wouldn't be able to guess how to run the project, and CI wouldn't work, and the deployment would fail as well.

app/feature/page_old_bak.tsx
lanterno, line 1: You don't need to commit a backup file ever in git, because we can always revert the commit and get the file back, or review the history as we want. 
Please remove this file

app/news/pageold_bak.tsx
lanterno, line 1: also this file is not needed. Please remove it.

app/page.tsx
```diff
@@ -10,6 +10,7 @@ export default function Home() {
   const [featureArticles, setFeatureArticles] = useState([])
 
   useEffect(() => {
+    console.log("Fetching publication on client...");
```
lanterno, line 13: Only use `console.log` only for local development, and don't push it to git.

utils/api.ts
```diff
@@ -64,25 +64,31 @@ export async function getPublications(params: {
...
   if (params.is_blog_post) searchParams.append('is_blog_post', 'true')
 
+  if (params.limit) searchParams.append('limit', params.limit.toString())
+  if (params.offset) searchParams.append('offset', params.offset.toString())
+
   const url = `${API_BASE_URL}/api/v1/media/publications/${
     searchParams.toString() ? `?${searchParams.toString()}` : ''
   }`
-
+  console.log('Fetching:', url)
```
lanterno, line 82: Remove console.log

lanterno:
I added some high-level comments from an organizational point for view. I'll leave the technical review to Ali. 

Overall, I think the PR is great, I like the idea of using a single component for all blogs, new, and featured articles. This is very nice. Reusing code is very important 👍

.gitignore
zixben: Yes, you're right — I'm aware of that. It was intentionally left out in this case, as it mainly contains rules for ignoring local files. If you'd like to include a .gitignore, we can simply use the one from the main branch — it's already set up the way it should be.

next.config.mjs
zixben: Yes, I'm aware of that as well — the same reasoning applies here. There's no guesswork needed, since the files are exactly the same as in the main branch. I only modified them locally to replace the image configuration with local ones to avoid unnecessary errors during development.

app/feature/page_old_bak.tsx
zixben: Yes, I'm aware of that — I only left it in to avoid any confusion during development. I can definitely remove it.

app/page.tsx
zixben: That one, in particular, was already there. Not by me!

utils/api.ts
zixben: This one was made by me. I will remove it. No worries.

components/paginated-publications.tsx
```diff
@@ -0,0 +1,150 @@
+// Shared PaginatedPublications Component for Blogs, News, and Feature
+
```
A-Allam, line 2: could you please remove this empty line

next.config.mjs
```diff
@@ -1,21 +1,34 @@
...
+// import MiniCssExtractPlugin from 'mini-css-extract-plugin'
 import { createProxyMiddleware } from 'http-proxy-middleware'
 
-const BACKEND_URL = process.env.NEXT_PUBLIC_API_URL || 'https://api.archetype.rancho.me'
+const BACKEND_URL = process.env.NEXT_PUBLIC_API_URL;
+console.log(" Backend URL:", BACKEND_URL); 
+// || 'https://api.archetype.rancho.me'
 
 const nextConfig = {
+  // images: {
```
A-Allam, line 9: why you removed this lines?

A-Allam (changes requested):

.gitignore
```diff
@@ -34,3 +34,6 @@ yarn-error.log*
 # typescript
 *.tsbuildinfo
 next-env.d.ts
+
+
+package-lock.json
```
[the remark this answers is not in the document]
zixben: I will include this as well.

app/blogs/page_Client-side.tsx
```diff
@@ -0,0 +1,136 @@
+'use client'
```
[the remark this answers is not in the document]
zixben: Yes, I have explained in the description at the top of the page. You can do whatever you want with that file.

app/blogs/page_SSR_NO-helper.tsx
```diff
@@ -0,0 +1,134 @@
+import Link from 'next/link'
```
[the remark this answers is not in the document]
zixben: Same here, I have explained in the description at the top of the page. You can do whatever you want with that file.

app/blogs/page_SSR_WITH-helper.tsx
```diff
@@ -0,0 +1,135 @@
+// Clean Helper-based Refactor (with shared logic)
```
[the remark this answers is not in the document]
zixben: Same here, I have explained in the description at the top of the page. You can do whatever you want with that file.

components/paginated-publications.tsx
zixben: I am sorry I don't understand what you ask me here.

next.config.mjs
zixben: I have explained to Ahmed as well. This is basically the reason I chose to ignore my local version of this file in the first place. To avoid throwing errors caused by having that enabled. I didn't need it in my local development environment.

zixben:
As I mentioned from the beginning, my intention was simply to share some parts I developed that could be attached or adapted to the main code. I have different local settings that don't need to be uploaded, and that’s why you might see some things commented out—those were causing issues in my local environment.

Please feel free to disregard those and use your own version as needed.

PR #17, opened 2025-05-27
Fixes/improve search facets

zixben:
- [x] [fix: cleaned and formatted the code](https://github.com/archetype-pal/frontend/commit/a01173e8fe38b24303037f25bab4e4ea38448bed)
- [x] [fix: Fixed the Radio Button Filter (for tow options filters)](https://github.com/archetype-pal/frontend/commit/253969cd30570d32fcf830ebed56f40e68fe8314)
- [x] [fix: in the process of fixing the Date filter](https://github.com/archetype-pal/frontend/commit/166cfcbe13dac715e292bef626de98674ff85161)

***
- [x] [fix(facets): pass baseFacetURL prop to facet components and improve f…](https://github.com/archetype-pal/frontend/pull/17/commits/f9bea09f340439e07ac90265eecccea144efc4d0)
- Improve DynamicFacets integration with backend API and fix missing baseFacetURL prop
#### Changes Made:
- [x] Added baseFacetURL prop to both range-based and categorical facet components in DynamicFacets.tsx.
- [x] Adjusted filter-config.ts to ensure correct renderConfig mapping per resultType and facet key.
- [x] Updated filter-order.ts to define facet rendering order based on backend-provided facet keys.
- [x] Improved DynamicFacets to handle missing or unexpected keys gracefully.
- [x] Updated the SearchPage.tsx to continue passing searchType to renderConfig for compatibility.
***


A-Allam (approved):
@zixben great work, and sorry for the delayed review!
Just a small suggestion — when the user clicks on a filter option from the side menu, the other options disappear. From a UX perspective, it might be better to keep all options visible and simply highlight the selected one.


https://github.com/user-attachments/assets/44214a86-1bb3-43a9-bce9-a4ff20d8b6db

zixben:
@A-Allam Thank you, and no worries! I completely agree with your point. However, this behavior follows the way [Models of Authority](https://www.modelsofauthority.ac.uk/digipal/search/facets/?hi_has_images=Without+image&page=1&img_is_public=1&locus=face&result_type=manuscripts&view=grid) is originally constructed, and it was asked to take that as an example and replicate its approach. Not particularly that, but the way the filters behave in general. It was actually more complex to implement it this way than to simply leave the other options visible and mark them as inactive(or something similar), but I aimed to stay consistent with the reference.

zixben:
@A-Allam  Also, I forgot to mention that the filters are rendered dynamically from the back end. As a result, unselected options are not hidden or being hidden—they simply aren’t present. This is because the API’s faceted response no longer includes them once the selection has already been narrowed.

PR #36, opened 2025-07-08
Fix filter reset on pagination and persist pagination controls

zixben:
- Ensure filtered query parameters persist when navigating pages
- Always show 'results per page' selector regardless of result count
- Keep Pagination component rendered at all times
- Conditionally hide page buttons if only one page exists

- [x] [Fix filter reset on pagination and persist pagination controls](https://github.com/archetype-pal/frontend/commit/9dcc2c3525ae9150810750a0d2c5d4139e360b5f)


lanterno (approved):
LGTM.. but let's wait a day for Ali's review and we merge.

A-Allam (approved):

PR #66, opened 2026-07-06
Feat/internationalization customization

geourjoa:
Add I18n translation system (Er + Fr for now) and a way to configure some site label (like the title, some footer, header texts ...) in the interface. It will allow user to configure the website to suit his implementation.

They will be always hardcoded text or bad translation in code for sure, but that is inevitable I think. I will do a second pass after this first merge.

The current customizable labels are saved in a plain file. Not ideal to replicate an instance. I work (in an other coming PR) to save them in DB.


lanterno (approved):
LGTM

PR #98, opened 2026-07-23
feat(backoffice): manuscript image upload UI (background uploads, resume)

saad-mhmd:
## What & why
The frontend half of image upload (archetype-pal/backend#72). Adds an "Add images"
flow to each item part in the backoffice, consuming the chunked-upload API.

## Key changes
- **Chunked upload service** (`services/backoffice/uploads.ts`): session API + an XHR
  chunk transport (fetch can't report upload progress) + an orchestrator that
  creates/resumes a session, sends only missing chunks, finalizes, and polls to a
  terminal state. Pure helpers (extension allowlist, filename→locus guess, chunk
  planning) are unit-tested.
- **Background uploads**: a shell-level manager + floating tray so uploads survive
  backoffice navigation; per-file progress, cancel, and **Retry** (server resume
  re-sends only missing chunks). The dialog is now a thin staging form.
- **Reload recovery**: localStorage breadcrumbs re-attach to conversions that finished
  during a reload, or prompt to re-select a file to resume (strict name+size match).
- **Two-tab safety**: a file already uploading in a live sibling tab is parked as
  'busy' rather than racing; chunk error bodies are sanitized (no Django HTML traceback
  ever reaches a toast).
- Also: one-shot "reindex in Search Engine" nudge after a batch; `.openapi.json`
  slice for the upload endpoints; edit-dialog save fixed to send `image` only when
  repointed to a non-empty path.

## Testing
Local CI-equivalent green: tsc, eslint, vitest, `next build`, bundle-budget (the
+gzip for reload-recovery is an intentional, refreshed budget). Manually clicked
through on the site and on the full infrastructure stack (upload → navigate away →
tray persists → thumbnail; duplicate; cancel; retry; reload/resume; two-tab).

Part of #72 — needs the backend `feat-72/item-image-upload` PR for its API.


lanterno (commented):
Re-reviewed at **b95d557**. This adds the whole backoffice image-upload surface: a chunked-upload service over XHR, a shell-level sequential queue that survives navigation, localStorage breadcrumbs for reload recovery, a two-tab ownership protocol, and a staging dialog + tray. The design is right and the hard parts — resume, re-attach-after-reload, cross-tab arbitration — are thought through rather than hand-waved. What blocks it is merge ordering, a cluster of cancel/failure paths that tell the editor an upload failed when it actually succeeded, and the i18n rule.

(The branch was force-pushed since my last pass; the code under review is unchanged, so everything below is against the current head.)

## What's good

- **`watchUploadSession` split out of `uploadImageFile`** (`services/backoffice/uploads.ts:244-276`) is the decomposition that makes reload recovery possible at all — once finalize returns, every byte is server-side and only the polling loop was lost. The comment at `:237-243` says exactly that.
- **`chunkErrorDetail`** (`uploads.ts:99-109`) refuses to surface an HTML body and caps plain text at 300 chars. Getting a Django debug traceback into a toast is the classic failure of upload UIs, and you pre-empted it deliberately.
- **`findSessionRival`** (`lib/backoffice/upload-breadcrumbs.ts:232-248`) is a real deterministic tie-break — `createdAt` then `id`, so exactly one side ever yields — not a hopeful heuristic. Tested in both directions.
- **The identity-preserving `setInterrupted`** (`contexts/upload-manager-context.tsx:577-582`). Non-obvious and correct: a fresh-but-equal array on a 30 s interval re-renders every consumer and can feed render→effect→scan loops.
- **Not hashing sha256 client-side** (`uploads.ts:298-301`), with the reason written down. Right call.
- **`upload-breadcrumbs.test.ts`** is the strongest suite here — malformed-payload rejection, the 50-entry cap, all three partition classes, and `findActiveDuplicate`'s negative cases. It asserts rather than smoke-runs.
- **Extracting `FloatingPanel`** and rewiring the search-engine task panel through it instead of copy-pasting the chrome.

---

## Blocker — merge order

### Merged alone, this ships a permanently broken "Add images" button
`components/backoffice/manuscripts/item-parts-tab.tsx:230`

The button renders unconditionally on every item part, and the flow targets `/api/v1/uploads/sessions/` (`services/backoffice/uploads.ts:56`). api `main` has no `apps/uploads` and no `uploads` route in `config/urls.py`; api #133 is still open (CI green, unmerged).

**Failure:** an editor stages a scan and clicks Upload. `createUploadSession` 404s. Django's URL resolver returns no JSON `detail`, so `describeUploadError` falls through to `Request failed (404).` — that string in the tray row, and a `Upload failed: f12r.tif` toast carrying it — with a Retry button that reproduces it forever, and nothing anywhere saying the feature isn't deployed.

Don't merge ahead of api #133. Gating the button behind a flag would be the wrong fix: the flag would have exactly one lifetime — until #133 lands — and would then need removing, so it buys nothing that merge ordering doesn't.

---

## Major

### 1. Cancelling during "processing" reports "Canceled" for an upload that succeeds — and Retry then hard-fails
`contexts/upload-manager-context.tsx:333` · `components/backoffice/uploads/upload-tray.tsx:200,278`

`upload-tray.tsx:200` folds `processing` into `active`, so `:278` routes the X to `onCancel`. `cancel` (`:392-403`) only calls `controller.abort()` — but by `processing`, `finalizeUploadSession` has returned and api pr/133 `apps/uploads/services.py:386` has already dispatched `ingest_upload.delay(...)`. The conversion completes regardless.

Two consequences in drain's AbortError branch (`:328-335`). It never calls `invalidateManuscript`, so the grid keeps the stale list — but so does `startWatch`'s abort branch, and that one is deliberate. The branch that is out of step is the crumb handling: every other terminal path removes the breadcrumb (`:324`, `:331`, `:340`, `:464`, `:472`); this one alone keeps it, rewritten to `status: 'canceled'` (`:334`) — which is what makes the reload behaviour below misfire.

Worth knowing: api pr/133 `services.py:392-394` **refuses** to abort a `PROCESSING` session at all. There is genuinely nothing to cancel at that point, so the control shouldn't claim to.

**Failure:** editor uploads `f12r.tif` (800 MB). Bytes finish, tray flips to "Converting to JP2…". They click X, tray says "Canceled." Celery publishes the image anyway. Retry then hits `_resolve_active_session_collision` (`services.py:160-162`, `resumable = status in (PENDING, UPLOADING)`) → `DestinationBusy` with code `session_active`, which `isConflictError` rejects → a hard red *"Another upload session is already targeting 'uploads/item-part-3/f12r.jp2'."* for an upload that succeeded.

### 2. One non-retryable poll failure marks a successful upload as failed
`services/backoffice/uploads.ts:262`

The poll loop does a bare `session = await getUploadSession(token, session.id)` with no failure tolerance. `fetchWithRetry` (`services/backoffice/api-client.ts:18-30`) retries GETs only on thrown network errors and 502/503/504 — three attempts over ~1.5 s. Anything else throws, lands in drain's generic else (`:342-348`), and produces status `error`, an `error` crumb, and a red toast, while the ingest task runs to completion.

**Failure:** a 900 MB TIFF converts for ~12 minutes (≈360 polls at the 2 s default). An API restart lasting more than ~1.5 s anywhere inside that window exhausts the three attempts. The editor is told "Upload failed: f12r.tif", clicks Retry, and gets the same 409 as above. The image exists and only appears after a reload.

### 3. Cancel never calls the abort endpoint — sessions and temp chunks are stranded
`services/backoffice/uploads.ts:56`

`uploads.ts` exports only create/get/finalize (`:62-76`); nothing in the branch issues a DELETE. The endpoint exists and is wired (api pr/133 `views.py:68-75` → `services.abort_session`, `services.py:392`, which rmtrees the temp dir and deletes the row), your own `.openapi.json` documents it as `upload-sessions-abort`, and `backofficeDelete` is already sitting at `api-client.ts:89-91`.

**Failure:** editor A cancels a 3 GB scan at 50%. 1.5 GB of chunk files stay under the session temp dir and the row stays in `pending`/`uploading` (both in `UploadSession.ACTIVE_STATUSES`, `models.py:55`). Editor B uploads the same scan and gets a bare red *"Another upload session is already targeting…"* with no affordance to clear it. Repeat across a cataloguing session and `_check_disk_space` (`services.py:207-212`) starts 507-ing on a disk full of abandoned bytes. `dismissInterrupted` (`upload-manager-context.tsx:610-614`) leaks the same way.

### 4. ~40 new user-visible strings are hardcoded English
`components/backoffice/manuscripts/item-image-upload-dialog.tsx:121`

The diff adds exactly two catalog keys (`manuscriptsDetail.addImages`, `manuscriptsDetail.noImagesYet`) — both consumed by `item-parts-tab.tsx`, which already had `useTranslations`. Everything net-new is hardcoded: the upload dialog imports no i18n at all (`:121`, `:122-126`, `:156`, `:240`, `:249`, both toasts at `:77`/`:111`, all aria-labels); `upload-tray.tsx` the same (`:52-57`, `:68`, `:76-77`, `:234`, `:242`, `:252`); `upload-manager-context.tsx:118, 237-243, 264, 348`; `uploads.ts:108, 227, 232`; `floating-panel.tsx:59`.

**Failure:** a French editor clicks the translated "Ajouter des images" and gets an entirely English dialog and tray for the rest of the flow — including the messages they most need, e.g. *"Server error (500) while uploading a chunk — you can retry the upload."*

<details>
<summary><b>Minor</b> (9)</summary>

**A cancelled upload comes back after a reload as "Interrupted by a reload"** — `contexts/upload-manager-context.tsx:334`, `:399`. Cancellation persists as `updateUploadBreadcrumb(id, { status: 'canceled' })` rather than removing the crumb, and `partitionUploadBreadcrumbs` (`upload-breadcrumbs.ts:174-187`) never consults `status` — only tabId and age. `getUploadTabId` reads sessionStorage, which survives a same-tab reload, so the crumb comes back own-tab adoptable; `routeSessionCrumb` finds the session pending/uploading and returns `'prompt'` (`:537`). *Failure:* editor cancels a 2 GB upload, reloads, and is asked to re-select the file to resume the upload they just cancelled — blaming a reload that had nothing to do with it. Reloading again brings it back.

**"The next scan or reload settles it" is false** — `contexts/upload-manager-context.tsx:474`. The item was added to `itemsRef` by `addItem` at `:440`, and `scanBreadcrumbs` skips any crumb whose item is live in this tab (`:562`). Nothing removes it from `itemsRef` except `dismiss`/`clearFinished`, which also delete the crumb. So neither the 30 s rescan nor the storage listener can re-settle it — only a full reload. drain's error branch (`:342-348`) has the same shape. This is what makes major 2 unrecoverable inside the tab rather than self-healing.

**`describeUploadError` degrades DRF field errors** — `services/backoffice/uploads.ts:224`. It reads only `err.body?.detail`, else `Request failed (${status}).`. api pr/133 `views.py:42-43` validates via `is_valid(raise_exception=True)`, so 400s are field-shaped with no `detail` key (only `services.UploadError` goes through `_error_response`, `views.py:27-31`). `lib/backoffice/format-api-error.ts` already flattens exactly those bodies and every other backoffice mutation uses it. Reachable: `serializers.py:17-18` caps locus at 72 and tags at 255, and the dialog's inputs (`item-image-upload-dialog.tsx:241-246`, `:250-255`) carry no `maxLength`. *Failure:* a >255-char tag list yields `Upload failed: f12r.tif — Request failed (400).` with no hint which field, forever.

**The `duplicate` branch is a dead end for an orphaned JP2** — `contexts/upload-manager-context.tsx:336`. `_check_destination_free` (`services.py:132-139`) raises `DestinationExists` for a file on disk *or* an existing ItemImage row. The file-without-row case is reachable from this UI: `deleteItemImage` (`item-image-edit-dialog.tsx:88`) removes only the row, and `apps/manuscripts/models.py` has no `post_delete` or `delete()` override that unlinks the file. `createItemImage` is exported at `services/backoffice/manuscripts.ts:75` with zero consumers, so there's no re-attach path either. *Failure:* delete a mis-catalogued image, re-upload the corrected scan under the same name → "Already present — A file already exists at…", the grid refreshes to nothing, and no retry/rename/Browse-Media can attach it. That filename+part is permanently blocked. This one needs a backend sub-code on `destination_exists` to distinguish the two cases, so it belongs with api #133 rather than here.

**Clearing the IIIF image path is silently ignored but reports success** — `components/backoffice/manuscripts/item-image-edit-dialog.tsx:74`. `if (trimmed && trimmed !== (image.image ?? '')) payload.image = trimmed;` drops `image` when the field is empty, while the field stays editable and the preview blanks. *Failure:* editor clears the path to unlink a wrongly-pointed image, watches the preview go blank, saves, gets a green "Image updated" — and the row still points at the old JP2.

**Queued uploads cannot be escaped** — `components/backoffice/uploads/upload-tray.tsx:279`. The X is `disabled` for `pending` and only routed to `onCancel` when active. Uploads are strictly sequential (`upload-manager-context.tsx:255`). *Failure:* drag-drop the wrong folder, stage 40 scans, click Upload. Items 2-40 sit behind a greyed-out X with no way to drop them; the only escape is a reload, which then turns the whole tail into "Interrupted by a reload" prompts.

**Tray and task panel overlap** — `components/backoffice/common/floating-panel.tsx:41` hardcodes `fixed bottom-4 right-4 z-50 w-[380px]`. `UploadTray` mounts in the shell after `<main>` (`backoffice-shell.tsx:69`); the search-engine `TaskProgressPanel` now renders through the same component with no positional override. Same stacking context, same z-index → DOM order wins and the tray covers the panel. The reindex nudge (`upload-manager-context.tsx:240-243`) routes the user to exactly that page, so the designed flow walks into the collision.

**Logout unmounts the manager without aborting** — `components/backoffice/layout/backoffice-shell.tsx:40`. The `!token` early return unmounts `UploadManagerProvider` the instant the token clears; the provider has no unmount cleanup, so nothing aborts `controllers.current`. The token is revoked, so the still-running XHR dies on a 401 mid-chunk and leaves a stranded session (see major 3). The `beforeunload` guard (`:620-628`) never covered in-app navigation anyway.

**Locus/Tags labels are not associated** — `components/backoffice/manuscripts/item-image-upload-dialog.tsx:240,249`. No `htmlFor`, no `id` on the inputs at `:241`/`:250`; `StagedFileItem` repeats per staged file, so N files give 2N unlabelled fields. The sibling dialog does it correctly (`item-image-edit-dialog.tsx:136-140`).

</details>

<details>
<summary><b>Nits</b> (7)</summary>

**The resume path has no cross-tab guard** — `lib/backoffice/upload-breadcrumbs.ts:216,242`. Both guards exclude same-id crumbs. `enqueue` mints a fresh uuid so they bite there, but `resumeInterrupted` (`upload-manager-context.tsx:589-608`) reuses the crumb id (`itemFromCrumb` sets `id: crumb.id`, `:150`) and re-stamps `tabId` on the one shared crumb — so two tabs resuming the same crumb are mutually invisible and both PUT into one session. Minting a fresh id on resume closes it, but that touches `promptOnlyRef`, `resumedIds` and the `setInterrupted` filter, so it's your call rather than something to slip in.

**`guessLocusFromFilename` invents plausible-wrong loci** — `lib/backoffice/upload-helpers.ts:41`. `/(\d{1,4})\s*([rv])\b/gi` has no left anchor, so on `Add_ch_33792r` the engine fails at the first digit, advances one character and matches `3792r` → `f.3792r`; `MS_12345v.tif` → `f.2345v`. Unlike a genuinely ambiguous name (which correctly yields `''`), these look legitimate. `(?<!\d)` fixes it and leaves `f12r`, `0032v` and `MS_Add_1234_f005r` unchanged.

**Raw response body `console.error`'d in production** — `services/backoffice/uploads.ts:142`. `chunkErrorDetail` exists precisely because that body can be a Django debug page; this line puts it in the console anyway. `lib/api-fetch.ts:16` has the house `isDev` gate.

**The reindex nudge can fire three times for one batch** — three uncoordinated counters call `showReindexNudge`: `:354`, `:484-486`, `:583`.

**Batches over 50 files silently lose breadcrumbs** — `lib/backoffice/upload-breadcrumbs.ts:61,109`. `writeAll` sorts by `updatedAt` desc and slices to 50; `enqueue` stamps an identical `now` on every crumb (`:363`, `:383`), and the heartbeat keeps live items at the top, so the evicted crumbs are exactly the not-yet-started ones. `updateUploadBreadcrumb` on a missing id is a documented no-op, so nothing errors. A quire is realistically >50 folios and the dialog imposes no limit.

**localStorage read-modify-write is not atomic across tabs** — every mutator (`:117-118`, `:122-131`, `:133-137`, `:141-146`) does list→transform→write-whole-array. The module docstring already frames breadcrumbs as best-effort, so this is informational; one key per crumb would fix it if it ever matters.

**Bare `crypto.randomUUID()`** — `upload-manager-context.tsx:366`, `item-image-upload-dialog.tsx:66`. It's secure-context-only and throws straight out of a drop handler on a plain-HTTP origin. Weaker than it looks: the repo guards it in `lib/search-query.ts:564`, `components/search/query-builder-panel.tsx:30` and `components/manuscript/manuscript-annotorious.tsx:62-63`, but `lib/saved-searches.ts:41` is already bare — so there's no single house pattern to appeal to.

</details>

## Test coverage

The four new suites assert properly rather than smoke-running, and `upload-breadcrumbs.test.ts` is genuinely good. But the two things the PR leads with are untested. `putChunk` (`uploads.ts:111-156`) has zero tests — progress reporting, the abort-listener add/remove pairing, non-2xx → `ChunkUploadError`. `uploadImageFile` (`:286-341`) is untested end to end, including the `sentBytes` anchor at `:314` and `if (!missing.has(chunk.index)) continue;` at `:317`. Invert that `continue`, or build `missing` from `received_chunks` instead of `missing_chunks`, and every test in this PR still passes; the bug surfaces only as a permanently stuck resume in production. In `upload-manager-context.test.tsx`, `uploadImageFile` is mocked wholesale, so drain's non-abort error branch, its 409 branch, `cancel`, and retry-after-error are all uncovered.

## Merge conflict

`.bundle-budget.json` is the only conflict (`git merge-tree origin/main pr/98` — messages/{en,fr}.json auto-merge cleanly), and it does **not** resolve in this branch's favour. main's `3087413` (30 Jul) raised the budget to `2145525` / last-measured `1950477` / 112 chunks. This branch's `00e799f` (22 Jul) sets `2141029` / `1946390` / 110 chunks, measured at `git_sha e124e78` — a commit the force-push deleted. Taking `--ours` would *lower* the CI budget below what main deliberately granted, on a measurement nobody can reproduce. Take main's file, then re-run `just bundle-budget-update` on the merged tree.

## What I'd do next

1. Hold the merge until api #133 lands — nothing else matters until then.
2. Make the three "success reported as failure" paths honest: don't offer cancel during `processing` (the backend refuses to abort it anyway — offer "stop tracking" instead), tolerate a handful of consecutive poll failures before calling a conversion dead, and call `DELETE /uploads/sessions/{id}/` on real cancels and dismissals so sessions and chunks aren't stranded.
3. Do the i18n sweep — a `backoffice.uploads.*` namespace in both catalogs, with `uploads.ts` returning codes the presentation layer translates.
4. Add the two tests that pin the headline behaviour: which chunk indices `uploadImageFile` actually PUTs for `missing_chunks: [1, 3]`, and the `sentBytes` sequence.

I'm opening a follow-up PR against `feat-72/image-upload-ui` with the small mechanical fixes — the abort endpoint, poll tolerance, the cancel/X routing, `formatApiError` delegation, the locus regex anchor, the dev-gated `console.error`, the label associations, and the panel offset. Take it, cherry-pick from it, or ignore it entirely. It deliberately leaves out the i18n sweep (the French copy should be yours), the transport tests, the resume id re-key, and the orphaned-JP2 dead end (that one needs a backend sub-code).

lanterno:
Reviewed at **6613749**. I went through the earlier review item by item against the code rather than the PR body, and ran the touched suites locally (7 files, 92 tests green; en/fr/de catalogues carry identical key sets and `i18n-messages.test.ts` passes; `git merge-tree` onto current `main` is clean, so the bundle-budget conflict is gone and the budget was re-measured on top of `main`, not below it).

## Earlier review: what's actually closed

Everything that was called a blocker or major is fixed, with the fix in the right place:

- Cancel during `processing`/finalizing is now "stop tracking" (`upload-tray.tsx:212-213,297`), and `cancelAll` excludes those phases (`upload-manager-context.tsx:546-551`) — matching the backend's `ABORTABLE_STATUSES`.
- The poll loop tolerates five consecutive failures (`uploads.ts:278,311-318`), with tests.
- `cancel`, `cancelAll` and `dismissInterrupted` all issue the DELETE (`ctx:516,566,796`).
- i18n is complete: 51 `backoffice.uploads.*` keys in three locales, `uploads.ts` returns codes and `errorText` translates them, and a grep of the four new components finds no literal UI text.
- The rest of the minors and nits are closed too: cancelled crumbs are removed rather than rewritten, `formatApiError` is delegated to, the empty IIIF path is refused with a translated message, pending items are cancellable and retry de-dups the queue, the task panel is offset while the tray has content, sign-out awaits `cancelAll()` before `logout()` and drain refuses a token change, labels are associated, the locus regex is anchored, the prod `console.error` is gated, the nudge fires once, `randomUUID` is guarded, and `uploads.test.ts:191-310` now pins which chunk indices a resume PUTs and the `sentBytes` sequence.

Left open and honestly listed under Known limitations: the resume path reusing the crumb id (`ctx:783-784`), the non-atomic localStorage read-modify-write, and the breadcrumb cap (now 500, still silent past it). Fine to carry.

**The merge-order blocker still applies** — the button renders unconditionally (`item-parts-tab.tsx:231`) and api#133 is open. Merge after it, as a merge commit, as the PR says.

## New — minor

**1. Retry after a processing-side give-up hard-fails an upload that is succeeding.** `ctx:578-590`, `uploads.ts:293,303-309`. When `watchUploadSession` gives up (the 30-minute `processTimeoutMs`, or six consecutive poll failures from an API outage of roughly ten seconds) the item goes `error` with Retry, and the crumb keeps its `sessionId`. `retry` re-runs `uploadImageFile` from scratch → `createUploadSession` → the backend finds the destination in `processing`, `resumable` is false → `session_active` 409 → the red "Another upload session is already targeting…" toast. A reload does the right thing via `routeSessionCrumb`; nothing inside the tab does. A 4 GB TIFF whose lossless conversion runs past 30 minutes is the realistic trigger. The same path is what the user hits if finalize itself times out at nginx on a large file (raised on backend#133). Fix: in `retry`, when the crumb has a `sessionId`, GET the session first and `startWatch` if it is assembled/processing/complete; only re-create when it is genuinely gone.

**2. `cancelAll` drops the crumb even when the server refused the abort.** `ctx:553-556` aborts the controller before the DELETE; the real `uploadImageFile` then rejects with `AbortError`, and drain's abort branch (`:395-402`) removes the crumb unconditionally, before the awaited DELETE settles. So the "drop only crumbs the server actually discarded" logic at `:573-575` only holds for items with no controller. The test at `ctx.test:667` passes because `uploadImageFile` is mocked and never rejects on abort. Effect: sign out during a transfer while the API is briefly unreachable → session survives server-side, crumb is gone, the sign-out dialog's promise isn't kept. Set a "cancelling" flag before aborting and have the abort branch skip crumb removal when it's set, or let `cancelAll` own the removal.

## Nits

- **No chunk-level retry.** `putChunk` (`uploads.ts:138-191`) turns one dropped socket in a 20-minute transfer into a red toast and a manual click. One or two automatic re-PUTs of the same index are safe — the server de-dups indices and rejects a wrong-size chunk.
- **`enqueue` writes crumbs one at a time** (`ctx:498-501`), each a full parse and stringify of the array; with the cap at 500 that's ~500 rewrites of up to 125 KB in the click handler. A batch `saveUploadBreadcrumbs(list)` fixes it.
- **Breadcrumbs are per-browser, not per-user.** User A's interrupted uploads are offered to user B on a shared workstation, and re-selecting yields a red `DestinationBusy`. Stamping the owner's user id on the crumb and filtering in `scanBreadcrumbs` avoids it.
- **Dismiss during `finalizing` can still toast "Upload failed"** (`ctx:592-599` vs `:427-436`): finalize isn't abortable, so if it then fails the generic branch toasts for an item the user just dismissed. Rare.

One backend-side note that affects what this UI shows: `create_session` checks the disk before it looks for an active session, and ingest writes the JP2 straight to its final path, so while a colleague's upload of the same filename is `processing` this UI reports "Already present" rather than the busy state. That's raised on backend#133; nothing to change here.

PR #105, opened 2026-08-04
fix(labels): wire model labels to the backend SiteLabel API

geourjoa:
## Pull request overview

Wires the frontend “model labels” system to read/write label values from the backend SiteLabel API (instead of a local JSON config), and updates the footer + backoffice translations UI to use the new footer label keys.

It's a PR to fix afterward internationalization stuff. Sorry for that. If not merged, all the translations stuff editable from backoffice will be not persistent on container restart.

**Changes:**
- Replace local JSON-based model-label persistence with backend fetch/PUT to `/api/v1/site-labels/` (with SSR-safe fallback to defaults).
- Rename footer-related label keys and update footer rendering to use the new lines.
- Update the backoffice translations page metadata and the internal API route to support the new write flow.

### Reviewed changes

Copilot reviewed 5 out of 5 changed files in this pull request and generated 2 comments.

<details>
<summary>Show a summary per file</summary>

| File | Description |
| ---- | ----------- |
| lib/model-labels.ts | Renames footer label keys and updates default localized footer strings. |
| lib/model-labels-server.ts | Implements backend-backed read/write for model labels via `apiFetch`/`authFetch`. |
| components/layout/footer.tsx | Switches footer content to render `footerLine1`, `footerLine2`, and `footerBottomLine`. |
| app/backoffice/translations/page.tsx | Updates backoffice field metadata to expose the new footer label keys. |
| app/api/model-labels/route.ts | Updates PUT to write labels via backend and adds error handling. |
</details>


lanterno (commented):
Re-verified at `4887dcb` against current `main` (`8f4b8c7`). It still merges cleanly (`git merge-tree`) even though `main` rewrote `components/layout/footer.tsx` for backend-driven Partners.

Moving model labels off the container-local JSON file and onto `/api/v1/site-labels/` is the right fix for a real bug — file-backed labels don't survive a restart and aren't visible to a second replica. The read path's caching is well judged. Two things block it:

1. The write path is a full-blob overwrite of all 22 keys aimed at a backend deliberately built for per-key upsert, reachable from a form that silently renders hardcoded defaults whenever the backend is unhealthy. That combination can wipe every stored label.
2. Merging as-is regresses the public footer to placeholder text, and no change in this repo can fix that.

## What's good

- **The diagnosis.** `config/model-labels.json` is written inside the container, so every admin edit is lost on restart and invisible to a second replica. This is the correct fix, not a workaround.
- **The endpoint actually exists.** `api` main has `SiteLabelsView` wired at `apps/common/urls.py:10`, the `SiteLabel.Key` enum, and migration `0008` seeding all 22 rows. Worth contrasting with the site-features PR `main` reverted in `bfef3d0`: `AppSettings` and `/api/v1/site-features/` were never shipped backend-side, which is why that one had to come out. This PR has no equivalent *code* gate — it does have a *data* gate, see blocker 2.
- **The key set matches exactly.** All 22 keys line up with `SiteLabel.Key.values`, so the backend's whole-payload unknown-key rejection (`apps/common/views.py:166-168`) never fires. Easy to get wrong.
- **`next: { revalidate: 60, tags: [SITE_LABELS_TAG] }` earns its place.** `readModelLabels` is awaited from `app/layout.tsx:57` and `:85` (including `generateMetadata`), the footer, `components/content/publication-pages.tsx`, and four `(site)` segments — an uncached fetch really is several round-trips per render.
- **Error propagation on write.** Attaching the upstream status to the thrown `Error` and mapping it back in the route (403 stays 403) is better than collapsing everything to 502.
- **`lib/pages-server.ts`.** Logging `normalizePage` returning `null` on a 200 closes a genuine blind spot — that was previously indistinguishable from a legitimate 404.

## Blockers

### 1. A degraded read is served as a healthy 200, and any save then writes all 22 keys

`lib/model-labels-server.ts:38,45,50` — `readModelLabels()` returns `getDefaultModelLabelsConfig()` on a non-2xx, on a malformed 200 body, and on a thrown fetch. `app/api/model-labels/route.ts:19-20` serves that as a plain `200`, so the backoffice `useQuery` resolves successfully and `app/backoffice/translations/page.tsx:230-234` runs `setConfig(serverConfig)` into a fully editable form. Nothing distinguishes "these are the stored labels" from "the backend was down, here are hardcoded defaults".

Saving round-trips through `writeModelLabels`, whose body is `JSON.stringify({ labels: normalizeModelLabels(config.labels) })` (`lib/model-labels-server.ts:59`). `normalizeModelLabels` iterates `Object.keys(DEFAULT_MODEL_LABELS)` (`lib/model-labels.ts:105`), so it **always** materialises all 22 keys — the payload is never a subset — into the backend's `update_or_create` loop at `apps/common/views.py:185-187`.

To be fair about what's new here: the base did the same two things. `readJsonConfig(CONFIG_FILE, …, defaults)` also fell back silently, and `writeModelLabels` also wrote the full normalized set. Under a file, that was harmless — if the file was unreadable, the defaults *were* the stored state, and a full write was the only option. Swapping the source for a network call is what makes the fallback *wrong* rather than *the truth*, and that is what turns the pre-existing full-set write destructive.

**Failure scenario:** the backend 500s or 429s for a few seconds. An admin opens `/backoffice/translations` in that window; the form renders the frontend defaults — visibly `siteTitle: "Archetype"` where the backend holds `"Models of Authority"`, and `footerLine1: "Footer first section"`. They fix one typo and hit Save. The now-healthy backend receives all 22 keys and `update_or_create`s every `SiteLabel` row to the frontend defaults. The toast says saved. It isn't recoverable from the audit log either: `SiteLabel` is registered via `register_audited_models` (`apps/common/apps.py:13`), and `on_save_handler` records `summary = str(instance)` — which `SiteLabel.__str__` defines as the bare key — with `payload=None`. Key names, no old value, no new value.

Two independent fixes, both needed:
1. Give `readModelLabels` a degraded signal so the GET route can answer `503` and the backoffice renders an error state instead of an editable form. (`/api/model-labels` has exactly one consumer — `app/backoffice/translations/page.tsx:145,154` — so a 503 there is safe.)
2. Send only the keys the admin actually changed.

### 2. Merging regresses the footer About column to placeholder text

`components/layout/footer.tsx:79-80`. The diff deletes `<p>{t('about')}</p>` and replaces it with `getLabel('footerLine1')` / `getLabel('footerLine2')`.

The frontend has never called `/api/v1/site-labels/` before this PR (`git grep site-labels origin/main` in this repo is empty), so the production `SiteLabel` rows hold exactly what migration `0008_sitelabel_per_key.py` seeded — and it seeds `footerLine1`/`footerLine2` to `"Footer first section"` / `"Footer second section"` (`"Pied de page, première section "` / `"…deuxième section "`, both with a trailing space). The frontend defaults at `lib/model-labels.ts:63-70` are the same placeholders, so there is no source of real copy on either side.

Meanwhile `messages/en.json` → `nav.footer.about` holds the actual copy — *"Scottish Charters and the Emergence of Government, 1100–1250. A resource for the study of the contents, script and physical appearance of the corpus of Scottish charters."* — plus its French translation. In the merged tree, `footer.tsx:29` is the only consumer of the `nav.footer` namespace and it no longer calls `t('about')`, so both become dead.

**Failure scenario:** this merges and deploys. Every page of the public site renders a footer About column reading `Archetype` / `Footer first section` / `Footer second section` where it previously carried the project description, and it stays that way until a superuser fills four fields across two locales in the backoffice.

This is the one thing I can't fix from this repo, so it's yours: either a data migration in `api` seeding `footerLine1`/`footerLine2` with the `nav.footer.about` text (and whatever the real funding statement is), or an agreed post-deploy step where an admin fills them in before the deploy is announced. Please don't delete `nav.footer.about` until one of those has landed — right now it's the only copy of that text in either repo.

## Major

### The write is a full-blob overwrite against a per-key upsert API — lost updates

`lib/model-labels-server.ts:59` and `app/api/model-labels/route.ts:46-48`. The backend is explicit about the contract (`apps/common/views.py:150-152`): PUT *"upserts only the keys present in the payload — unlike the old singleton's full-blob overwrite, keys absent from the payload are left untouched."* The same convention is already encoded in-repo on current `main` at `app/api/site-features/route.ts:44-51`.

**Failure scenario:** Admin A opens `/backoffice/translations`. Admin B renames `appManuscripts` to "Corpus" and saves. A, whose form still holds the pre-rename snapshot, later edits an unrelated field and saves — A's PUT includes `appManuscripts` and silently reverts B's rename. No conflict, no diff, and per the audit note above, nothing in `EditEvent` but a key name.

Note the route re-normalizes the client body at `:46-48`, so a client-side diff alone won't fix this; the route has to stop materialising absent keys too. Fixing this also removes the latent coupling where adding one frontend-only key would 400 the *entire* payload.

## Minor

**`app/api/model-labels/route.ts:61-67` — the justification for `'minutes'` is wrong, and the call marks stale rather than purging.** The second argument is a cacheLife profile, but what the runtime consumes from it is `expire`, not `revalidate` — so `'minutes'` doesn't "match the `next.revalidate: 60`", it sets the tag's expiry an hour out. Traced through the installed `next@16.2.10`: `server/revalidation-utils.js:121-123` → `durations = { expire: cacheLife.expire }`, and `'minutes'` is `expire: 3600` (`server/config-shared.js:147-151`); `server/lib/incremental-cache/file-system-cache.js:62-64` then writes `stale = now` and `expired = now + 3600000`, where the no-profile branch at `:69-73` writes `expired: now`.

In fairness to the outcome: this is not a no-op. `stale = now` makes `areTagsStale` true (`tags-manifest.external.js:44-52`), `incremental-cache/index.js:375` marks the entry stale, and `patch-fetch.js:756` foreground-revalidates during static generation, while `:759-777` serves the stale value once and refreshes in the background on a dynamic render. So an edit does show up — on the next render, or the one after. It just isn't the immediate purge the comment claims, and it leaves a spurious hard-expiry an hour later.

`revalidateTag(SITE_LABELS_TAG, { expire: 0 })` is what the comment describes: it takes the object branch, yields `durations = { expire: 0 }`, writes `expired: now`, and satisfies `areTagsExpired` (`incremental-cache/index.js:373`) for a real cache miss. Type-valid — `revalidate.d.ts` declares `profile: string | CacheLifeConfig` where `CacheLifeConfig = { expire?: number }`. (`updateTag` isn't an option: `revalidate.js:53-54` throws when `workStore.page` ends with `/route`.)

**`lib/model-labels.ts:71-80` — the `footerBottomLine` default replaces a rights notice with a product blurb.** Migration `0008`'s header says it *"Mirrors DEFAULT_MODEL_LABELS in the frontend's lib/model-labels.ts. Kept in sync manually"*, and seeds `footerBottomLine` to `"©2015–17 Models of Authority. Some parts available under CC-BY licence. All manuscript images are copyright of their respective repositories. Website by DDH / KDL. Built with Archetype."` The frontend defaults it to three sentences describing what the Archetype software is. So the moment the backend is briefly unreachable, the site's copyright and image-rights notice is replaced by a product description — attribution disappears exactly when things are already going wrong. (`siteTitle` and `siteTagline` also diverge; those read as deliberate generic platform defaults, so it's the migration's "kept in sync" claim that overstates things there, not the code.)

**`app/api/model-labels/route.ts:51-60` — the detailed upstream error is built, then thrown away.** `writeModelLabels` reads `await res.text()` and embeds it in the `Error` message; the catch reads only `err.status` and returns a fixed `{ error: 'Failed to update site labels' }`. `err.message` is never returned and never logged. In development the `isDev` branch in `apiFetch` wins, so a failed PUT prints as `[API] OK PUT /api/v1/site-labels/ → 400`. An admin hitting the backend's per-key validation sees "Failed to save", and a developer reproducing it sees a line tagged `OK` — the reason has to be dug out of the Django log.

**Three doc comments describe events that never happened.** `git grep site-labels origin/main` returns nothing and the pre-image of `model-labels-server.ts` reads `readJsonConfig('model-labels.json', …)`, so the frontend never called `/api/v1/site-labels/` before this PR. That makes all of these fiction: `lib/model-labels-server.ts:26-30` ("the absence of any `next.revalidate`/`tags` here made that route's `revalidatePath` call a no-op and meant every single page render round-tripped to the backend"), `lib/model-labels-server.ts:41-43` and `lib/pages-server.ts:16-20` ("the `/site-labels/` 429 that prompted this logging pass"), and `lib/api-fetch.ts:8-11` ("exactly how a backend rate-limit throttling `/site-labels/` went unnoticed in production until someone grepped the raw access log by hand"). The PR description states the real before-state correctly — please just say that instead; as written it will mislead the next person debugging caching here.

**No tests at all, in a repo that tests the directly analogous module.** `lib/site-features-server.test.ts` and `app/api/site-features/route.test.ts` both exist on this branch; there is no `lib/model-labels-server.test.ts` and no `app/api/model-labels/route.test.ts`. `lib/model-labels.test.ts` never referenced `footerFunded`/`footerCopyright`, which is why renaming three members of a public key union produced zero test churn and a green CI. Untested: the defaults fallback per failure mode; the status-carrying thrown error; the route's status mapping and 502 default; and whether the read is actually tagged — drop the `next` init in a later refactor and `revalidateTag` silently stops mattering, with lint, tsc, vitest and the bundle-budget gate all still green. The fetch-mocking templates you want already exist in history at `387ac61`.

<details>
<summary><b>Nits</b> (6)</summary>

- **`lib/api-fetch.ts:33-39`** — the new non-2xx `console.error` sits in an `else if`, so `isDev` wins first: in development a 500 logs via `console.log` tagged `OK`/`SLOW`, i.e. the one environment where a developer is watching is where failures are least visible. Deriving the tag from `res.ok` as well as duration fixes both halves. (Logging non-2xx unconditionally also means production browsers log a `400` on a mistyped password, since 13 `'use client'` modules import this — I think that's an acceptable trade, but it's a deliberate one.)
- **`lib/api-fetch.ts:42-47`** — the catch is now unconditional, but an `AbortError` is not a failure. `services/tei-ref-search.ts:53` forwards an `AbortSignal` into `apiFetch`, driven by `components/backoffice/tei-ref-picker.tsx:211-216` via `useQuery({ queryKey: ['tei-ref-search', segment, query], queryFn: ({ signal }) => … })` — TanStack Query v5 aborts the in-flight request on every key change, i.e. every keystroke. Typing "Melrose" emits six `FAILED … AbortError` console errors in production, burying the genuine failures this change exists to surface. `app/(site)/hands/[id]/hand-viewer.tsx:106` is on the same path.
- **`lib/model-labels.ts:77-78`** — French stray space before a comma: `'…de rechercher' + ' , de visualiser…'` renders `de rechercher , de visualiser`. French takes a narrow no-break space before `; : ! ?`, never before a comma. (Moot if `footerBottomLine` is reconciled with the backend seed.)
- **`app/api/model-labels/route.ts:69`** — the route returns its own normalized config rather than the backend's PUT response, which re-reads and returns `{'labels': rows}` after the atomic upsert (`apps/common/views.py:189-190`). That response goes straight into the query cache via `queryClient.setQueryData` (`app/backoffice/translations/page.tsx:242`). The sibling route does the opposite and documents why (`app/api/site-features/route.ts:58-59`). This stops being optional once the write becomes a diff — the route has to return the full config from somewhere.
- **`app/api/model-labels/route.ts:30`** — `verifySuperuser` is now a redundant second round-trip: `SiteLabelsView` declares `permission_classes = [IsSuperuserOrReadOnly]` (`apps/common/views.py:155`), so the write is already gated. If `/api/v1/auth/profile` is what's being throttled, it returns non-ok and the save fails with `403 Superuser access required` — a misleading authorization error for a rate-limit condition, which is the exact class of silent degradation this PR set out to kill. Not introduced here; just newly redundant.
- **`config/model-labels.json`** — nothing reads it after this change (`git grep json-config-file` hits only `lib/site-features-server.ts` and the two site-features test files), yet it's still committed, still listed at `.prettierignore:25`, and lines 79-86 are the only surviving occurrence of `footerFunded`/`footerCopyright` in the tree. A contributor will edit it, see no effect, and lose time.
- **`app/backoffice/translations/page.tsx:33-47`** — the three new/renamed field metas hardcode English `title`/`description` in a backoffice that is otherwise fully bilingual (the same component resolves section headings via `t('translations.…')`). Pre-existing across all 22 entries, so not this PR's fault — flagging it so it doesn't get further entrenched.

</details>

## What I'd do next

1. Make the write a diff. `handleSave` sends only keys whose `en`/`fr` differ from `serverConfig`; the route stops re-materialising absent keys; `writeModelLabels` returns the backend's `{labels}` response so the client cache stays whole. That single change kills the lost-update bug, caps the blast radius of a degraded read, and removes the add-a-key-and-everything-400s coupling.
2. Give `readModelLabels` a `degraded` flag, `503` from the GET route, and an error state in the backoffice instead of an editable form full of defaults.
3. Reconcile `footerBottomLine`'s default with the migration's rights notice (which also removes the French stray space). Leave `footerLine1`/`footerLine2` mirroring the seed — inventing frontend-only copy there wouldn't change what production renders and would just add a new divergence. The real fix is the `api` seed in blocker 2, and `nav.footer.about` should stay put until that lands.
4. `revalidateTag(SITE_LABELS_TAG, { expire: 0 })`, and cut the comment down to one accurate line.
5. Rewrite the three doc comments to describe the actual before-state (the PR description already gets it right).

I'm opening a follow-up PR against `fix/wire-model-labels-to-backend` with items 1-5 plus the api-fetch logging fixes, the orphaned config file, and the two missing test files — take it, cherry-pick from it, or ignore it entirely. Blocker 2 is the one I've deliberately left alone, because it needs the real copy and a decision from you.

Issue #11, opened 2025-04-09
Search - Improve manuscript search facets


lanterno:
Each type of of search object has a list of available facets

<img width="300" alt="Image" src="https://github.com/user-attachments/assets/bd58fa21-4304-4ddd-b867-6e380d2d91e2" />

For the manuscript object type, there's a list of facets that are available here: https://api.archetype.gla.ac.uk/api/v1/search/item-parts/facets

for each of these facets, we need to display a filter on the page like how it looks now (or better take a look at how it used to work before): 
https://www.modelsofauthority.ac.uk/digipal/search/facets/?page=1&result_type=images&img_is_public=1&view=grid&locus=face

Issue #10, opened 2025-04-09
Search - move result types to be a bar at the top of the search page


lanterno:
Currently, the search page only searches for manuscripts, but it should be able to search for multiple types of objects. 

For now, we will keep it at only one type -> Manuscripts. However, we will display the other types while saying page under construction. 

Here's a list of the different search object types:

- Manuscripts
- Images
- Text
- Graphs
- Clauses
- Places
- People

Here's how it looked on the old page: 

<img width="487" alt="Image" src="https://github.com/user-attachments/assets/84ecd140-1a71-4d4b-92a8-67279158ae58" />

After Ali's work, now it looks like this on the new site:

<img width="380" alt="Image" src="https://github.com/user-attachments/assets/9198e036-1e06-4d9f-9a11-45f5339e22ba" />

However, since this is not really a filter as it changes the type of the object itself. Please put it at the top of the page: 

something like this:

<img width="1000" alt="Image" src="https://github.com/user-attachments/assets/8fad0108-9053-4a07-bd56-c00911652750" />

PR #108, opened 2026-08-05
feat(backoffice): add user impersonation UI for admins

geourjoa:
## Summary
Cross-repo dependency: requires backend PR archetype-pal/backend#152 (`POST /api/v1/auth/management/users/{id}/impersonate/`, superuser-only, token-swap) to work end-to-end.

There was previously no frontend UI for impersonation at all. Adds:

- An "Impersonate" action on the backoffice users table (`app/backoffice/users/page.tsx`), disabled with an explanatory tooltip for the current user's own row and for staff/superuser targets (mirrors the backend's own restrictions), gated behind a confirmation dialog.
- `startImpersonation`/`stopImpersonation`/`isImpersonating` on the auth context (`contexts/auth-context.tsx`): stashes the admin's real token in a second cookie while impersonating, restores it on "stop".
- A site-wide `ImpersonationBanner` (mounted in the root layout, so it's visible on both the public site and backoffice — impersonation targets are always non-staff, so the session lands on the public site after switching) with a "Stop impersonating" button.
- `services/backoffice/users.ts`: `impersonateUser(token, id)`.

## Test plan
- [x] `vitest run` — full suite 117 files / 1171 tests passed (19 new)
- [x] `tsc --noEmit` clean
- [x] `eslint` / `prettier --check` clean
- [ ] Smoke test against the real backend once #152 is merged

No new dependencies.

---
Side note (unrelated, left untouched): noticed `messages/en.json`'s `login.signIn` key holds French text ("Se connecter") in the English catalogue — pre-existing, out of scope here.

### DIFF
--- a/app/backoffice/users/page.test.tsx
+++ b/app/backoffice/users/page.test.tsx
@@ -1,0 +1,166 @@
+import * as React from 'react';
+import { render, screen, within, fireEvent, waitFor } from '@testing-library/react';
+import { QueryClient, QueryClientProvider } from '@tanstack/react-query';
+import { describe, it, expect, vi, beforeEach } from 'vitest';
+
+import type { PaginatedResponse, UserListItem } from '@/types/backoffice';
+
+const push = vi.fn();
+vi.mock('next/navigation', () => ({ useRouter: () => ({ push }) }));
+
+const startImpersonation = vi.fn();
+let mockAuthUser: { id: number; username: string } | null = { id: 1, username: 'admin' };
+vi.mock('@/contexts/auth-context', () => ({
+  useAuth: () => ({
+    token: 'admin-token',
+    user: mockAuthUser,
+    startImpersonation,
+  }),
+}));
+
+vi.mock('sonner', () => ({ toast: { success: vi.fn(), error: vi.fn() } }));
+
+const getUsers = vi.fn();
+const createUser = vi.fn();
+const updateUser = vi.fn();
+const deleteUser = vi.fn();
+const impersonateUser = vi.fn();
+vi.mock('@/services/backoffice/users', () => ({
+  getUsers: (...args: unknown[]) => getUsers(...args),
+  createUser: (...args: unknown[]) => createUser(...args),
+  updateUser: (...args: unknown[]) => updateUser(...args),
+  deleteUser: (...args: unknown[]) => deleteUser(...args),
+  impersonateUser: (...args: unknown[]) => impersonateUser(...args),
+}));
+
+import { toast } from 'sonner';
+import { TooltipProvider } from '@/components/ui/tooltip';
+import UsersPage from './page';
+
+function baseUser(overrides: Partial<UserListItem>): UserListItem {
+  return {
+    id: 0,
+    username: '',
+    email: '',
+    first_name: '',
+    last_name: '',
+    is_staff: false,
+    is_superuser: false,
+    is_active: true,
+    date_joined: '2024-01-01T00:00:00Z',
+    last_login: null,
+    ...overrides,
+  };
+}
+
+const ADMIN = baseUser({ id: 1, username: 'admin' });
+const STAFF = baseUser({ id: 2, username: 'staffer', is_staff: true });
+const SUPERUSER = baseUser({ id: 3, username: 'superadmin', is_superuser: true });
+const REGULAR = baseUser({ id: 4, username: 'regular' });
+
+function usersResponse(results: UserListItem[]): PaginatedResponse<UserListItem> {
+  return { count: results.length, next: null, previous: null, results };
+}
+
+function renderPage() {
+  const qc = new QueryClient({ defaultOptions: { queries: { retry: false } } });
+  return render(
+    <QueryClientProvider client={qc}>
+      <TooltipProvider>
+        <UsersPage />
+      </TooltipProvider>
+    </QueryClientProvider>
+  );
+}
+
+function rowFor(username: string): HTMLElement {
+  return screen.getByText(username).closest('tr') as HTMLElement;
+}
+
+beforeEach(() => {
+  push.mockClear();
+  startImpersonation.mockClear();
+  vi.mocked(toast.success).mockClear();
+  vi.mocked(toast.error).mockClear();
+  getUsers.mockReset();
+  createUser.mockReset();
+  updateUser.mockReset();
+  deleteUser.mockReset();
+  impersonateUser.mockReset();
+  mockAuthUser = { id: 1, username: 'admin' };
+  getUsers.mockResolvedValue(usersResponse([ADMIN, STAFF, SUPERUSER, REGULAR]));
+});
+
+describe('UsersPage impersonation action', () => {
+  it("disables the impersonate button for the signed-in admin's own row", async () => {
+    renderPage();
+    await screen.findByText('admin');
+
+    const button = within(rowFor('admin')).getByLabelText(/impersonate user/i);
+    expect((button as HTMLButtonElement).disabled).toBe(true);
+  });
+
+  it('disables the impersonate button for staff rows', async () => {
+    renderPage();
+    await screen.findByText('staffer');
+
+    const button = within(rowFor('staffer')).getByLabelText(/impersonate user/i);
+    expect((button as HTMLButtonElement).disabled).toBe(true);
+  });
+
+  it('disables the impersonate button for superuser rows', async () => {
+    renderPage();
+    await screen.findByText('superadmin');
+
+    const button = within(rowFor('superadmin')).getByLabelText(/impersonate user/i);
+    expect((button as HTMLButtonElement).disabled).toBe(true);
+  });
+
+  it('enables the impersonate button for a regular, non-staff, non-self row', async () => {
+    renderPage();
+    await screen.findByText('regular');
+
+    const button = within(rowFor('regular')).getByLabelText(/impersonate user/i);
+    expect((button as HTMLButtonElement).disabled).toBe(false);
+  });
+
+  it('opens a confirmation dialog naming the target user when clicked', async () => {
+    renderPage();
+    await screen.findByText('regular');
+
+    fireEvent.click(within(rowFor('regular')).getByLabelText(/impersonate user/i));
+
+    expect(await screen.findByText(/impersonate "regular"/i)).toBeTruthy();
+  });
+
+  it('on confirm, calls impersonateUser, then startImpersonation + navigates home on success', async () => {
+    impersonateUser.mockResolvedValue({ auth_token: 'target-token' });
+    renderPage();
+    await screen.findByText('regular');
+
+    fireEvent.click(within(rowFor('regular')).getByLabelText(/impersonate user/i));
+    await screen.findByText(/impersonate "regular"/i);
+
+    fireEvent.click(screen.getByRole('button', { name: /^impersonate$/i }));
+
+    await waitFor(() => expect(impersonateUser).toHaveBeenCalledWith('admin-token', 4));
+    await waitFor(() => expect(startImpersonation).toHaveBeenCalledWith('target-token'));
+    expect(push).toHaveBeenCalledWith('/');
+    expect(toast.success).toHaveBeenCalled();
+  });
+
+  it('shows an error toast and does not impersonate on failure', async () => {
+    impersonateUser.mockRejectedValue(new Error('boom'));
+    renderPage();
+    await screen.findByText('regular');
+
+    fireEvent.click(within(rowFor('regular')).getByLabelText(/impersonate user/i));
+    await screen.findByText(/impersonate "regular"/i);
+
+    fireEvent.click(screen.getByRole('button', { name: /^impersonate$/i }));
+
+    await waitFor(() => expect(toast.error).toHaveBeenCalled());
+    expect(startImpersonation).not.toHaveBeenCalled();
+    expect(push).not.toHaveBeenCalled();
+  });
+});
--- a/app/backoffice/users/page.tsx
+++ b/app/backoffice/users/page.tsx
@@ -2,6 +2,7 @@
 
 import { useMemo, useState } from 'react';
 import { useTranslations } from 'next-intl';
+import { useRouter } from 'next/navigation';
 import { useQuery, useMutation, useQueryClient } from '@tanstack/react-query';
 import { useAuth } from '@/contexts/auth-context';
 import {
@@ -12,6 +13,7 @@ import {
   Plus,
   Pencil,
   Trash2,
+  VenetianMask,
   Eye,
   EyeOff,
   CheckCircle,
@@ -50,7 +52,13 @@ import {
   BackofficeErrorState,
   BackofficeLoadingState,
 } from '@/components/backoffice/common/query-state';
-import { getUsers, createUser, updateUser, deleteUser } from '@/services/backoffice/users';
+import {
+  getUsers,
+  createUser,
+  updateUser,
+  deleteUser,
+  impersonateUser,
+} from '@/services/backoffice/users';
 import { backofficeKeys } from '@/lib/backoffice/query-keys';
 import { formatApiError } from '@/lib/backoffice/format-api-error';
 import { runBulkAction } from '@/lib/backoffice/bulk-action';
@@ -68,6 +76,13 @@ function getInitials(user: UserListItem): string {
 
 function fullName(user: UserListItem): string {
   return [user.first_name, user.last_name].filter(Boolean).join(' ');
+}
+
+// Mirrors the backend's own restrictions (apps.users.services.impersonate_user):
+// never yourself, never a staff/superuser account. No point offering an action
+// that is guaranteed to 400/403.
+function canImpersonate(row: UserListItem, currentUserId: number | undefined): boolean {
+  return row.id !== currentUserId && !row.is_staff && !row.is_superuser;
 }
 
 function relativeTime(dateStr: string | null, t: (key: string) => string): string {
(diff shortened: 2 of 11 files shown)


lanterno (commented):
Adds the frontend half of superuser impersonation: a per-row action on the backoffice users table, `startImpersonation`/`stopImpersonation` on the auth context backed by a second cookie that stashes the admin's token, and a site-wide amber banner. The mechanism is well chosen and the tests are real. The problem is the session model layered on top of backend #152's "hand the caller the target's own persistent `Token`" contract: one blocker — signing out while impersonating revokes the *target's* credential, not yours — and three separate routes to ending up wearing another identity with no banner.

Branch base is 19 commits behind `main`. The only files both sides touch are `messages/en.json` and `messages/fr.json`, and the hunks are far apart (`main` around ~898/~1504/~1678, this branch at ~1840/~1878/~2562), so the merge should be textually clean.

## What's good

- **The nested-impersonation guard** (`contexts/auth-context.tsx:52`). Stashing only when not already impersonating — so a second hop can't overwrite the real admin token with an already-impersonated one — is the non-obvious correct call, and it has a dedicated test.
- **The generation guard gets reused for free.** The profile effect's `active` flag (`auth-context.tsx:103-133`) means a fetch in flight for the old identity can't set a profile for the new one.
- **`canImpersonate` mirrors the backend's own rules** (`app/backoffice/users/page.tsx:84`) rather than letting the admin discover them via a 400/403, with a distinct reason string per disabled case.
- **The tests assert rather than smoke-run.** `app/backoffice/users/page.test.tsx:163-164` checks that `startImpersonation` was *not* called and `push` did not fire on failure — the assertion people usually skip — and `contexts/auth-context.test.tsx` drives real `document.cookie` instead of mocking the cookie module.
- **`handleStop` calls `router.refresh()`** (`components/impersonation-banner.tsx:24-28`). Right instinct about Next's client Router Cache under an identity change — which is exactly why its absence on the start path stands out below.

---

## Blocker

### 1. `logout()` while impersonating revokes the target user's token and spares yours — `contexts/auth-context.tsx:75`

```ts
if (token) {
  void logoutUser(token).catch(() => {});   // ← `token` is the TARGET's while impersonating
}
setAuthToken(null);
clearImpersonatorTokenCookie();             // ← admin's token discarded locally, never revoked
```

`token` is whatever is currently active, and `startImpersonation` set it to the target's token at line 56. `utils/api.ts:104` POSTs `/api/v1/auth/token/logout`; Djoser's `TokenDestroyView` deletes the *authenticated* user's token. Backend #152 `apps/users/services.py:34` is `token, _ = Token.objects.get_or_create(user=target)` — the target's own persistent `Token` row, not a scoped credential. So this destroys a third party's credential. Line 80 then throws away the admin's superuser token client-side with no revoke at all — the precise hazard the comment on lines 71-73 says the revoke exists to prevent.

**Failure:** admin impersonates `jdoe`, gets pushed to `/`, and clicks the sign-out icon in the public header (`components/layout/header.tsx:425`, `onClick={logout}`, inside the `{token ? … }` branch at line 406 — so it's on screen throughout impersonation, roughly one control away from the banner's Stop button). Django deletes `Token(user=jdoe)`. On his own machine, holding a 30-day `archetype_auth_token` cookie, jdoe starts 401ing on his next request and is signed out of a session he never touched. Meanwhile the admin's superuser token is still valid server-side, just no longer in anyone's hands.

**Fix:** revoke the stashed impersonator token, never the impersonated one — `const revokeToken = getImpersonatorTokenCookie() ?? token;`. Two lines. And add `expect(logoutUser).toHaveBeenCalledWith('admin-token')` to the existing logout test (`contexts/auth-context.test.tsx:137-155`) — it currently passes while the wrong token is revoked.

---

## Major

### 2. `stopImpersonation` hides the banner but leaves you as the target when the stash is missing — `contexts/auth-context.tsx:61-68`

`setIsImpersonating(false)` on line 67 is outside the `if (originalToken)` block. With no stash the function does nothing except remove the only indicator that you are not yourself: `token` and `archetype_auth_token` still hold the target's token, and the way back is gone. This is the default behaviour on the unhappy path, and it is untested — `contexts/auth-context.test.tsx:96-113` covers only the stash-present case.

**Fix:** make it fail hard, and keep it navigation-free. Clear the stash first, then `setAuthToken(originalToken)` unconditionally — `setAuthToken(null)` already clears the auth cookie, so a missing stash becomes a full local sign-out and `handleStop`'s existing `push('/')` drops you on the public home as a guest. Being signed out beats silently wearing someone else's identity. (Don't push `/login` from inside `stopImpersonation`: `handleStop` calls `router.push('/')` on the very next line and would override it.)

### 3. The stash expires while the auth cookie is renewed forever — `lib/auth-token-cookie.ts:43-50`

Both cookies get `Max-Age = 30 days` (lines 2 and 8) but only one is ever rewritten. `setAuthTokenCookie` is reached from `setAuthToken`, which the mount effect calls on *every page load* (`auth-context.tsx:93`), so `archetype_auth_token` rolls forward indefinitely for an active user. `setImpersonatorTokenCookie` has exactly one call site — `startImpersonation` (`auth-context.tsx:53`) — and is never called again. The stash therefore carries a hard deadline fixed at the moment impersonation began and is *guaranteed* to die first. Under Safari/ITP, which caps script-written cookies at 7 days, that's inside a week.

**Failure:** admin impersonates `jdoe` on 1 June and doesn't click Stop. They keep using the site daily, so jdoe's token is refreshed to a new 30-day window on every load. The stash expires on 1 July (or 8 June in Safari). On the next page load the mount effect finds no stash, sets `isImpersonating = false`, and the banner never renders — while the active token is still jdoe's. The admin now browses *and writes* as jdoe indefinitely, with no indicator and no recovery path. Given this project's authorship model (Graph has no author FK; the actor lives in `EditEvent`), every edit from then on is attributed to jdoe with nothing to distinguish it.

### 4. The impersonated session persists 30 days and auto-resumes on the next visit — `lib/auth-token-cookie.ts:8`

The mount effect at `auth-context.tsx:89-90` deliberately rehydrates the flag, so a browser reopened days later comes back up mid-impersonation. There's no expiry, no auto-stop, and `stopImpersonation` never tells the server anything — the token it drops is the target's real long-lived `Token` row, so it stays valid regardless. Two live credentials, one of them a superuser's, sit in JS-readable (non-`HttpOnly`) cookies for a month, and the feature's whole point is to park the admin on the public site where the rest of the app's XSS surface lives.

**Fix for both 3 and 4 is the same and it's small:** make the impersonation pair **session cookies**. Drop `Max-Age` from `setImpersonatorTokenCookie`, and have `setAuthTokenCookie` omit `Max-Age` when `getImpersonatorTokenCookie()` is non-null — both functions live in the same module, so that's one conditional and no new parameter, and no call site can forget it. Ordering already works in your favour: `startImpersonation` writes the stash before calling `setAuthToken`, and the fix in #2 clears it before calling `setAuthToken`. Closing the browser then ends impersonation cleanly, both cookies die together, and the asymmetry that produces #3 stops existing. The tradeoff is that an admin who impersonates and then closes the browser has to log in again; that seems clearly correct for this feature.

---

## Minor

### 5. `canImpersonate` ignores `is_active` — impersonating a deactivated user silently logs the admin out — `app/backoffice/users/page.tsx:84-86`

The predicate checks only self / staff / superuser. `UserListItem` carries `is_active`, the table renders an Active/Inactive column and offers bulk deactivate, so inactive rows are routinely on screen with the button enabled. Backend #152 has no `is_active` guard either (`impersonate_user` checks self at line 29 and staff/superuser at line 31, nothing else) and `Token.objects.get_or_create` happily returns a token for an inactive user — but DRF's `TokenAuthentication.authenticate_credentials` raises `AuthenticationFailed('User inactive or deleted.')`.

**Failure:** admin deactivates `jdoe`, then clicks Impersonate on his row to check what he sees. The API 200s, `startImpersonation` swaps the token in, `router.push('/')` runs, `GET /api/v1/auth/profile` 401s, and the catch at `auth-context.tsx:116-122` runs `setAuthToken(null); setUser(null)` — no toast, no navigation. The admin is now unauthenticated app-wide, looking at an amber banner reading "Viewing the site as " with a blank name. (Stop still works — the stash is intact — so it's recoverable, just baffling.)

**Fix:** `&& row.is_active` in `canImpersonate`, plus a third `reasonKey` branch and a `tooltipImpersonateInactive` key in both catalogs. Worth asking backend #152 for the matching service-side guard so the API can't hand out an unusable token.

### 6. Cross-tab: the cookie is browser-wide, the banner is per-tab — `contexts/auth-context.tsx:47`

`startImpersonation` writes the target token into the shared `archetype_auth_token` cookie, but `isImpersonating` is React state set only at that tab's mount (line 90) or by that tab's own call. There is no `storage` listener and no `BroadcastChannel` anywhere in the tree — the only `storage` listener is `hooks/search/use-search-view-mode.ts:40`. A second already-open tab keeps `isImpersonating === false` and its admin token in memory, while every server-side read in that tab now sees the target's cookie: `proxy.ts:101` (gates `/backoffice`) and `app/(site)/search/[type]/page.tsx:16` (derives `isResearcher`).

**Failure:** admin has the public site open in tab B, impersonates from the backoffice in tab A. Tab B shows the full admin chrome (backoffice link via `user?.is_superuser` at `header.tsx:408`, sign-out button) and no banner, but any SSR navigation there now carries jdoe's cookie, so the tab renders a mix of admin- and target-authenticated content with nothing indicating a switch. Clicking sign-out in tab B then revokes the in-memory admin token and clears both cookies, killing tab A's impersonation and its recovery stash — landing tab A in finding #2.

### 7. Start path omits `router.refresh()` — `app/backoffice/users/page.tsx:393`

`handleStop` gets this right; `impersonateMut.onSuccess` (lines 389-394) doesn't. Next's client Router Cache holds RSC payloads for routes already visited and restores them on back/forward regardless of `staleTimes` — and `next.config.mjs`'s `experimental` block contains only `optimizePackageImports`, so there's no override softening it.

**Failure:** admin visits `/search/graphs` (SSR under the admin cookie), goes to `/backoffice/users`, impersonates, lands on `/`, then navigates back to `/search/graphs`. The cached payload rendered under the admin's identity is served while the client half of the app is jdoe — invisible, because the banner asserts the session is jdoe's. One line. Note that `app/backoffice/users/page.test.tsx:9` mocks `useRouter: () => ({ push })` with no `refresh`, so the mock has to grow with the fix or the suite throws.

### 8. The fixed banner covers the sticky site header — `components/impersonation-banner.tsx:33`

`fixed inset-x-0 top-0 z-[100]` reserves no space and nothing else offsets for it, so on every public page the banner lands directly on top of the site header (`header.tsx:317`, `sticky top-0 z-50`) — the site title and nav are underneath it. The repo already publishes the mechanism for exactly this: `header.tsx:55-68` sets `--site-header-h` on `documentElement` "so other sticky chrome can offset below it", and the banner participates in neither direction.

I don't have a one-liner for you, and I'd flag the obvious ones as traps before you try them:

- **Dropping `fixed`** (in-flow banner) breaks the backoffice — `components/backoffice/layout/backoffice-shell.tsx:51` renders the entire shell as `fixed inset-0 z-50` (its loading state at :40 too), so an in-flow banner would be completely covered there. The `z-[100]` is load-bearing.
- **In-flow or `sticky`** also perturbs three viewport-height layouts that don't know about it: `app/(site)/layout.tsx:9` (`min-h-screen`), `components/search/search-page.tsx:57` and `:262`, and `app/(site)/search/loading.tsx:10` (all `calc(100dvh - var(--site-header-h))`).

So the honest options are (a) publish the banner's height as a CSS variable the way the header already does, and offset the header's sticky `top` plus those `100dvh` calcs, or (b) accept that the banner scrolls away. Both are real changes and both are your call — I've left this one out of my follow-up branch rather than guess.

Two smaller things in the same area: `z-[100]` also beats the Radix dialog overlay and content (both `z-50`), so the amber bar floats over any modal opened while impersonating — arguably correct for a safety indicator, but worth being deliberate about. And the banner shares the top strip with the root `<Toaster position="top-center">` (`app/layout.tsx:111`), including this feature's own "Now browsing as {username}" toast, which lands on top of it.

### 9. Banner shows the *admin's* username during the profile swap — `components/impersonation-banner.tsx:38`

`isImpersonating` flips synchronously inside `startImpersonation` while `user` is only replaced once the profile effect finishes fetching under the new token. Neither start nor stop clears `user`, so the banner mounts immediately and renders `t('bannerMessage', { username: user?.username ?? '' })` against the stale admin profile. On a slow connection it reads "Viewing the site as **admin**" for the whole round-trip — an actively false statement about the session's identity — and if the fetch fails (see #5) it degrades to "Viewing the site as " and stays there. The same window exists in reverse after Stop, where the header evaluates `user?.is_superuser` against the target's profile.

**Fix:** `setUser(null)` in both, and render a neutral string when `user` is null rather than interpolating an empty one.

### 10. French tooltip inverts the rule it states — `messages/fr.json:1842`

> `"Impossible d'emprunter l'identité d'un compte personnel ou superutilisateur"`

In that position *personnel* reads as the adjective **personal**, not the noun **staff**. The English source (`en.json:1842`) is "Staff and superuser accounts cannot be impersonated", so the French says the opposite: personal (i.e. ordinary, non-staff) accounts are exactly the ones that *can* be impersonated. A French-locale superuser hovers the disabled button on a staff row, reads that they can't impersonate a personal account, and concludes the feature is broken. Fix: `"…d'un compte du personnel ou superutilisateur"`. (`fr.json:1857` has the same slip in `permStaffLabel`, "Accès personnel" — pre-existing, not yours, mentioning it only so you know I saw it.)

### 11. "This action is audited" overclaims — `messages/en.json:1879` / `fr.json:1879`

Backend #152 writes exactly one `EditEvent` (`action=IMPERSONATED`, `services.py:36-42`) at token-issue time. Nothing done *during* impersonation is distinguishable from the target's own activity: requests carry the target's real token and `AuditActorMixin` (`apps/common/views.py:22-41`) binds `request.user` as the actor — i.e. the target. `stopImpersonation` never notifies the server, so there's no end marker and no way to bound the window. Given this project's authorship model, an impersonated edit is indistinguishable from a genuine one, while the dialog tells the admin the action is audited.

Either soften the copy to what is true ("The start of this impersonation is recorded in the audit log"), or make it true backend-side (stamp `impersonated_by` onto `EditEvent`s created under an impersonation token, and POST a stop event).

---

<details>
<summary><b>Nits</b> (4)</summary>

- **Query cache isn't reset on identity change** — `contexts/auth-context.tsx:47`. Neither start, stop nor `logout` touches the query client, and no key in `lib/backoffice/query-keys.ts` carries an identity component (the users list is `backofficeKeys.users.all()`, token only in `queryFn`, `page.tsx:222-223`). `browserQueryClient` is a module singleton (`components/providers/query-provider.tsx:17`) with `staleTime: 30 * 1000`. Narrow window: impersonate then hit browser-back before the profile refetch resolves, and the cached account roster is served to a session the server now treats as unprivileged. The same gap pre-exists on `logout`, so this is arguably a separate fix.
- **Disabled-button tooltip is keyboard-unreachable; banner uses `role="alert"` for permanent chrome** — `app/backoffice/users/page.tsx:769`. The disabled Button (line 774) is wrapped in a bare `<span>` — the standard pointer-events workaround — with no `tabIndex={0}`, and disabled buttons aren't tabbable either, so keyboard/SR users get a greyed icon with no route to the explanation the `t(reasonKey)` branch at line 783 exists to provide. Separately, `Alert` hardcodes `role="alert"` (`components/ui/alert.tsx:26`) — an assertive live region, wrong for status chrome. It spreads `{...props}` after `role`, so `role="status"` on the banner overrides it cleanly.
- **`IMPERSONATOR_TOKEN_COOKIE` export has no consumers** — `lib/auth-token-cookie.ts:69`. The sibling `AUTH_TOKEN_COOKIE` earns its keep (`proxy.ts:2,101`; `app/(site)/search/[type]/page.tsx:7,16`); this one is symmetry for its own sake and implies a server-side consumer that doesn't exist. Drop it until something needs it.
- **IIFE inside JSX** — `app/backoffice/users/page.tsx:760-787`. Every sibling action in the same `flex items-center gap-1` div is a plain `<Tooltip>` (line 788 onward); the IIFE exists only to hold two consts. Hoisting them into the row-map callback above its `return` would do. Related: `row.id !== currentUserId` (line 85) can't be the deciding clause while the management users viewset is superuser-only (`BasePrivilegedViewSet` → `IsSuperuser`, so `!row.is_superuser` already excludes your own row), and `currentUserId` is `user?.id` (line 761), which is `undefined` while the profile is in flight — the clause degenerates to `row.id !== undefined`, a permission-shaped predicate failing open. No reachable failure today; it goes live the moment the viewset is relaxed to staff.

</details>

<details>
<summary><b>Test-coverage gaps</b></summary>

The 16 new tests are genuinely good, but the suite is silent on every unhappy path above:

- `stopImpersonation` with a **missing** stash — the #2 branch. Only the happy path (`contexts/auth-context.test.tsx:96-113`) is covered.
- The logout test (`:137-155`) asserts the stash cookie is cleared but never **which token** `logoutUser` received. One line — `expect(logoutUser).toHaveBeenCalledWith('admin-token')` — would have caught the blocker outright; as written it passes while the wrong user's credential is revoked.
- Impersonating an inactive user: `baseUser` (`app/backoffice/users/page.test.tsx:40-54`) hardcodes `is_active: true` and no fixture overrides it, so #5 is invisible to the suite.
- The banner's stale-username window: `mockAuth.user` is set to the target directly, so the real sequencing (flag flips before the profile refetches) is never exercised.
- `services/backoffice/users.ts::impersonateUser` is mocked in the only test that touches it — nothing asserts the URL `${USERS_BASE_PATH}${id}/impersonate/` or the empty POST body.
- `app/layout.tsx` is changed but untested; nothing asserts the banner is mounted, or mounted inside `AuthProvider`.
- No en/fr key-parity check exists anywhere in the repo, which is why the fr mistranslation sails through.

</details>

---

## What I'd do next

1. Fix the blocker first — it's two lines in `logout()` and it's the one that damages a third party.
2. Take the cookie-lifetime pair together: session-scope both impersonation cookies (one conditional in `lib/auth-token-cookie.ts`) and make the missing-stash branch of `stopImpersonation` a hard local sign-out. That single change retires #2, #3 and #4 and removes the whole "silently wearing someone else's identity" class.
3. Add `is_active` to `canImpersonate` and `router.refresh()` to the start path — one clause and one line, plus a `refresh` in the page test's router mock.
4. The copy fixes (#10, #11) are free and both are currently telling admins something untrue.
5. Ask backend #152 for an `is_active` guard, and decide whether the "audited" claim gets softened or made real — that's a cross-repo call, not something to settle here.

I'm opening a follow-up PR against `feat/impersonation-backoffice-button` with 1-4 plus the regression tests, so you can merge it, cherry-pick from it, or ignore it entirely. It deliberately leaves the banner layout (#8), the cross-tab sync (#6), the query-cache reset (which is a pre-existing `logout` gap too) and the IIFE/`canImpersonate` cleanups to you — those are design calls rather than defects.

geourjoa:
I closed for know 
There is not so much use cases for know (lot of things are saved locally in browser...) and I fear to open door if not carefully released 

We will get back when it will be the time !

PR #109, opened 2026-08-05
feat(backoffice): Admin - Trash page for annotations

saad-mhmd:
Two things: the Admin → Trash page, and the viewer changes that go with soft delete.

**Admin → Trash:**  lists trashed annotations with row and bulk restore / permanent delete, filterable by type, who deleted it, and when. EN + FR.

**Viewer:**  deleting a text region now trashes it instead of hard-deleting: it calls the annotation `DELETE` rather than `unlink-region`, so the `corresp` survives and a restore brings the region and its text link back together. A trashed region keeps its `corresp`, so spans pointing at a region that is no longer on the image render as plain prose - no hover, no activate, no link styling. That also fixes refs left dangling by hard deletes predating the trash.

Delete confirms and toasts are localised EN + FR, and now say that an admin can restore it.

Needs archetype-pal/backend#157. Part of archetype-pal/backend#144.


app/backoffice/annotations/page.tsx
```diff
@@ -182,7 +182,7 @@ export default function AnnotationsPage() {
   const deleteMut = useMutation({
     mutationFn: (id: number) => deleteGraph(token!, id),
     onSuccess: () => {
-      toast.success('Graph annotation deleted');
+      toast.success('Graph annotation moved to trash');
```
geourjoa: Use i18n, translations are inside messages/fr.json and messages/en.json

I have missed some when I work on internationaliztion

app/backoffice/annotations/page.tsx
```diff
@@ -207,7 +207,7 @@ export default function AnnotationsPage() {
           action: (id) => deleteGraph(token!, Number(id)),
           invalidate: () =>
             queryClient.invalidateQueries({ queryKey: backofficeKeys.graphs.all() }),
-          pastTense: 'deleted',
+          pastTense: 'moved to trash',
```
geourjoa: Use i18n, translations are inside messages/fr.json and messages/en.json

I have missed some when I work on internationaliztion

components/manuscript/annotation-gallery.tsx
```diff
@@ -342,14 +342,14 @@ export function AnnotationGallery({
       if (!token) return;
       if (
         typeof window !== 'undefined' &&
-        !window.confirm(`Delete graph #${graphId}? This cannot be undone.`)
+        !window.confirm(`Delete graph #${graphId}? It moves to the trash; an admin can restore it.`)
```
geourjoa: Use i18n, translations are inside messages/fr.json and messages/en.json

I have missed some when I work on internationaliztion

components/manuscript/annotation-gallery.tsx
```diff
@@ -342,14 +342,14 @@ export function AnnotationGallery({
...
       ) {
         return;
       }
       setDeletedIds((prev) => new Set(prev).add(graphId));
       try {
         await deleteViewerAnnotation(token, graphId);
-        toast.success(`Graph #${graphId} deleted`);
+        toast.success(`Graph #${graphId} moved to trash`);
```
geourjoa: Use i18n, translations are inside messages/fr.json and messages/en.json

I have missed some when I work on internationaliztion

lanterno (commented):
20 files, +1087/-114. CI green across build, lint, format, test and lighthouse.

The central claim holds. I traced `deleteViewerAnnotation` → `DELETE /api/v1/annotations/graphs/{id}/` → `GraphViewerWriteViewSet`, which carries `TrashableViewSetMixin`; its `perform_destroy` is a `save()`, so `pre_delete` never fires and the `corresp` survives. That is exactly what makes the restore lossless, and the dangling-link handling that follows from it is well designed — the `Set | null` three-state, with `null` meaning "don't know" rather than "nothing is live", is the right call and the test that pins it is the most valuable one in the PR.

Six things below. Items 1 and 2 are the ones I would want resolved before merge.

---

## 1. The confirm dialog still promises a permanent delete

`app/backoffice/annotations/page.tsx:330` renders `annotations.deleteDesc`:

> "Graph #{id} ({allograph}) will be **permanently deleted**." / "sera **définitivement** supprimé."

That mutation is now a soft delete, and the success toast two lines above it was changed *in this PR* to "moved to trash". So the admin is told the annotation is gone forever, confirms, and is then told it is recoverable.

It is also asymmetric with `trash.purgeDesc`, which correctly reserves that phrase for purge — so within one feature "permanently deleted" now means two different things.

Every other user-facing string in this PR got the honest treatment ("It moves to the trash; an admin can restore it"). This dialog was the one that got missed, and it is the one that matters most, because it is the text someone reads immediately before committing to the action.

## 2. Deleting a linked text region silently dropped from superuser to any logged-in user

`trashSelectedRegion` swaps the endpoint, and the two endpoints do not carry the same permission:

|  | endpoint | viewset | permission |
|---|---|---|---|
| before | `POST .../image-texts/{id}/unlink-region/` | `ImageTextManagementViewSet(FilterablePrivilegedViewSet)` | `IsSuperuser` |
| after | `DELETE /api/v1/annotations/graphs/{id}/` | `GraphViewerWriteViewSet` | `IsAuthenticated` |

And `resolveViewerRole` (`lib/manuscript-viewer-access.ts:19`) returns `'editor'` for *any* authenticated user, which grants `canDeleteAnnotations: true`. So a plain logged-in account that previously got a 403 on "Remove region" now succeeds.

The soft delete bounds the blast radius, and it does bring TEXT regions in line with image graphs, which already went through the `IsAuthenticated` endpoint. So this may well be intended. But it is not in the description, and a change to who can destroy data deserves an explicit sentence — ideally in the PR body, so it is on the record for whoever audits this later.

## 3. The error toast stayed hard-coded, and its translation key is now orphaned

`app/backoffice/annotations/page.tsx:190`:

```ts
onError: (err) => { toast.error('Failed to delete annotation', { … }); }
```

Directly beneath the success toast this PR localised. `annotations.toastFailedDelete` already exists in both catalogs with exactly that English string. After this PR it and `toastDeleted` are both unreferenced — four dead entries across en/fr.

## 4. `runBulkAction` lost its compiler guarantee

`pastTense` / `noun` went from required to optional with `= 'updated'` / `= 'item'`. A caller that forgets both now silently ships "3 items updated" instead of failing to compile.

A discriminated union — `{ pastTense; noun } | { messages }` — keeps both call styles working and keeps the compiler enforcing that exactly one is supplied.

## 5. Stale bulk selection after a single-row restore

`tableEpoch` is bumped after both bulk actions but not after `restoreMut` / `purgeMut`.

Selection is keyed by real graph id (DataTable's default `getRowId` falls back to `row.id`), so this cannot purge the wrong row — I checked. But `selectedIds` reads the raw `rowSelection` map rather than the row model, so: select A and B → row-restore A → bulk Purge sends A's id → the backend 404s it, since purge targets the trash → "1 succeeded, 1 failed". Harmless, but confusing. Bumping the epoch in both mutations' `onSuccess` covers it.

## 6. Two nits

- The trash page builds the query params twice — once in the `filters` memo for the key, once inline in `queryFn`. Build once and pass it to both, so they cannot drift.
- `columns` memo deps are `[t, restoreMut]`; `useMutation` returns a fresh object every render, so the memo never actually holds. `restoreMut.mutate` is stable — depend on that.

---

## Checked and clear

Recording these because several are non-obvious and I did not want the author to think they went unexamined:

- **`liveGraphIds` really does see text regions.** All three `buildInitialViewerAnnotations` call sites pass `includeText: true`, and `saveAll`'s refresh re-fetches `'text'` explicitly — so the post-save `setInitialA9sAnnots(outcome.seed)` cannot blank the set and deaden every span. This was the failure mode I most expected to find, and it is not there.
- **The specificity comment is accurate.** globals.css `.viewer-text-panel [data-graph-id]:hover` is (0,3,0); the injected `:hover` variant ties on specificity and wins on document order.
- The `/^[\d,\s]+$/` guard before interpolating attribute values into CSS is good practice and worth keeping.
- `PAGE_SIZE = 50` is within `BoundedLimitOffsetPagination.max_limit` (100), so server pagination is not silently truncating.
- `graphs.all()` prefixes both `graphs.list()` and the trash-actors key, so one `invalidateGraphs()` refreshes all three queries.
- The viewer route target `/manuscripts/[id]/images/[imageId]` exists.
- en/fr are at exact parity, 2408 keys each.

And two that look like findings but are pre-existing rather than introduced here:

- Client-side `getSortedRowModel` under server pagination means a sort only reorders the loaded page. True of all four server-paginated backoffice pages, not something this PR introduced.
- The sidebar gates the Admin group on `is_staff` while those endpoints are `IsSuperuser`, so Users / Search Engine / Data Quality already over-show. Trash's placement is consistent with the group it joins.

PR #111, opened 2026-08-12
fix(site-features): wire site features to the backend AppSettings API

geourjoa:
## Summary
Cross-repo dependency: requires backend PR archetype-pal/backend#149 (adds `GET/PUT /api/v1/site-features/`, superuser-gated write, backed by the new `AppSettings` model) to actually work end-to-end — this PR alone is safe to merge first (it degrades to `getDefaultConfig()` on any fetch failure) but won't reflect real backend data until #149 is also merged.

- `config/site-features.json` was being read/written directly from the Next.js server's local filesystem — unsafe across replicas/deploys. Follows the exact same pattern already used for model labels (see `fix/wire-model-labels-to-backend`, commit 63f557c).
- `lib/site-features-server.ts`: `readSiteFeatures`/`writeSiteFeatures` now call the backend via `apiFetch`/`authFetch` instead of `readJsonConfig`/`writeJsonConfig`. All existing defensive merge/normalize logic (missing `features` key, malformed sections, etc.) is preserved, just re-pointed at the backend response instead of file contents.
- `app/api/site-features/route.ts`: PUT now forwards the superuser token to `writeSiteFeatures` and returns 502 on backend write failure.
- `config/site-features.json` is intentionally **not deleted** — same caution as `config/model-labels.json` (don't remove until every environment has migrated).

No downstream changes needed: `contexts/site-features-context.tsx` and the backoffice UI consume the internal `/api/site-features` route, whose contract to its own callers is unchanged.

## Test plan
- [x] `vitest run` — full suite 114 files / 1155 tests passed
- [x] `tsc --noEmit` clean
- [x] `eslint` / `prettier --check` clean
- [ ] Smoke test against the real backend once #149 is merged

No new dependencies.


lanterno (commented):
Moves site-features off the Next server's local filesystem onto the backend `AppSettings` API, puts a 60s tagged fetch cache in front of the read, and tightens the write gate to `is_superuser`. The wiring is right. Two things stop it merging as-is: the backend endpoint it calls isn't merged yet, and putting the read on the network makes the "fall back to `getDefaultConfig()`" path reachable for the first time — which turns the backoffice editor into a config-eraser.

## What's good

- The defensive read logic survived the transport swap **intact** rather than being rewritten — the `isPlainObject` guards, the `mergeFeatureFlags` backward-compat story for configs with no `features` key, and `normalizeSectionOrder` all carried across unchanged.
- `lib/site-features-server.test.ts` is a real rewrite, not a port. The read path now has a full failure matrix (non-ok, network throw, `[null, [], 'oops', 7]` bodies, `res.json()` rejecting, unknown/non-boolean flags, a partial `searchCategories` override that asserts sibling keys keep defaults *and* that the result isn't the defaults object itself). The write path asserts the exact path, token, method, headers and body reaching `authFetch`.
- Replacing the `mkdtemp` + `process.cwd()` scratch-directory harness in the route test with a small in-memory stand-in for the AppSettings rows is a strict improvement — no filesystem state, and it reads like the thing it models.
- Adding the cache tag + `revalidate: 60` in the *same* PR that moves the read onto the network was the right call — without it `app/layout.tsx:84` round-trips to the backend on every render.

**Fixed since my last pass:** `lib/site-features-server.ts:49-54` now logs `[API] GET /api/v1/app-settings/ → 200 with unexpected body shape` before falling back. That closes the larger half of the silent-degradation gap I flagged. (The `res.json()`-rejects half — truncated body, HTML error page — still falls through the unlogged `catch` at `:87`.)

---

## Blocking

*(2 and 3 are one root cause seen from two sides; the same change fixes both.)*

### 1. `/api/v1/app-settings/` does not exist yet — merging first is a visible regression

`lib/site-features-server.ts:10`. api#149 is still **open** (`gh pr view 149` → `state: OPEN, mergedAt: null`; api main is at `67cdbe1` and `apps/common/urls.py` has no `app-settings` route). Until it lands, every read 404s, `res.ok` is false, and line 44 returns `getDefaultConfig()`.

The PR body says this is "safe to merge first (it degrades to `getDefaultConfig()`)". It degrades without crashing, but `getDefaultConfig()` is **not** what the site serves today. The pre-PR read comes from `config/site-features.json`, and that file really is in the production image: `Dockerfile:59` copies `/app/.next/standalone`, and `.next/standalone/config/site-features.json` is present in the built output.

**Failure:** deploy with #149 still open and the public nav order changes from `search, lightbox, collection, blogs, featureArticles, about, news, events` (`config/site-features.json:12-21`) to `search, collection, lightbox, news, blogs, featureArticles, events, about` (`ALL_SECTION_KEYS`) — Lightbox and My Collection swap, News and About move. Simultaneously the backoffice editor goes inert: the PUT hits the same 404, throws at `lib/site-features-server.ts:119`, and `route.ts:62` answers 502, so every save shows *"Failed to save — Failed to update site features"* with nothing in the logs.

Merge #149 first (its `0010_seed_site_features.py` seeds the rows, so the order is right the moment it migrates), or land them together. Heads-up while you're sequencing: #149 also touches `apps/common/schema.yaml`, and open api#170 rewrites that file too.

### 2. A degraded read is served to the editor as a `200`, and the next save destroys the stored config

`lib/site-features-server.ts:44` / `:54` / `:87` → `app/api/app-settings/route.ts:18-21` → `app/backoffice/site-features/page.tsx:27-28,134-140`.

`readSiteFeatures()` collapses every failure into `getDefaultConfig()` with no way for a caller to tell a real config from a fallback, and the GET route returns it as a 200. The editor only throws on `!res.ok` and its render still branches on `isLoading` alone, so a fallback config renders as a normal editable form. `saveSiteFeatures` then PUTs the **whole** object (`page.tsx:32-49`), `writeSiteFeatures` replaces all four top-level keys (`:102-113`), and api#149's `SiteFeaturesView.put` runs `AppSettings.objects.filter(key__startswith=…).exclude(key__in=keys).delete()` before upserting.

Before this PR the fallback was effectively unreachable — the config file shipped in the image and always parsed. Moving the read onto the network makes it reachable on any transient fault.

**Failure:** the API container restarts during a deploy. A superuser opens `/backoffice/site-features` in that 20s window and sees every section enabled, every flag on, every column and facet at its default, in the default order — indistinguishable from a real config, with no error indicator. They flip one toggle and save. Every configured `visibleColumns`/`visibleFacets` selection, every disabled section and the custom `sectionOrder` are deleted. `readSiteFeatures` backfills defaults so the public site looks healthy and nobody notices until someone reopens the backoffice.

### 3. The fallback is all-permissive, so a blip re-enables everything an admin turned off

`lib/site-features-server.ts:39`. `getDefaultConfig()` (`lib/site-features.ts:134-157`) sets every section `true`, every flag `true`, every search category `enabled: true`. There's no last-known-good — line 39 rebuilds `defaults` fresh each call. `proxy.ts:114` gates on `sections[key] === false` and its own fallback is `{ sections: {}, searchCategories: {} }` (`proxy.ts:84`), so the middleware fails open too.

**Failure:** an admin has disabled the Lightbox section and turned off `manuscriptDescriptions` while TEI descriptions are mid-migration. `/api/v1/app-settings/` 500s for 30 seconds. In that window Lightbox is back in the nav and routable, `msDescEnabled` evaluates true, half-migrated descriptions render publicly — then it all snaps back silently when the backend recovers.

---

## Worth fixing before merge

### 4. `revalidateTag(TAG, 'minutes')` marks the entry stale, it does not purge it — and the comment says the opposite

`app/api/app-settings/route.ts:71`. Verified against the installed `next@16.2.10`:

- `revalidation-utils.js:121` → `durations = { expire: cacheLife.expire }`
- `config-shared.js:147-151` → `minutes = { stale: 300, revalidate: 60, expire: 3600 }`
- `file-system-cache.js:62-64` → writes `{ stale: now, expired: now + 3_600_000 }`
- `tags-manifest.external.js:28-43` → expiry requires `expiredAt <= now`, so only `areTagsStale` fires
- `patch-fetch.js:759-777` → a stale entry serves the **cached** body and revalidates in the background

Next's own code says as much: `revalidate.js:208-212` sets `pathWasRevalidated` only when there is no profile *or* `cacheLife.expire === 0`.

The immediate invalidation you get today comes entirely from the pre-existing `revalidatePath('/', 'layout')` on the next line: that resolves to `_N_T_/layout` (`revalidate.js:81-101`), which with no profile writes `{expired: now}`, and it is an implicit tag on the entry — `patch-fetch.js:715` passes implicit tags as `softTags`, `incremental-cache/index.js:337-340` folds them into `combinedTags`, and `areTagsExpired` at `:373` returns `null` (hard miss).

So the comment at `:65-70` calling `revalidateTag` "the actual fix" has it backwards, and it's an invitation to delete the line that is actually doing the work. Use `revalidateTag(SITE_FEATURES_TAG, { expire: 0 })` — which does purge, and typechecks (`revalidate.d.ts:9` accepts `string | { expire?: number }`) — shorten the comment, and **keep `revalidatePath`**. (`updateTag` is not an option: `revalidate.js:49-58` throws in route handlers.)

### 5. The PUT's pre-write read is now cache-mediated, and only the implicit `/layout` tag keeps it fresh

`app/api/app-settings/route.ts:52`. Before this PR, `const current = await readSiteFeatures()` read the file directly and was always current. It now goes through the same tagged fetch entry as public renders, and it is the source of `current.features` for the merge at `:58` — a read immediately before a destructive replace.

In today's single-process deployment this happens to stay correct, but only because of the mechanism in #4: the previous PUT's `revalidatePath('/', 'layout')` hard-expired the entry. Two ways it stops being correct — someone edits the rows out of band (Django admin, a script, the seed migration), or the server runs more than one process/replica, since `tagsManifest` is in-memory per process. Then line 52 can merge over a body up to 60s old, or over `getDefaultConfig()` if the read failed, and `mergeFeatureFlags(current.features, undefined)` silently re-enables a flag an admin just disabled — the exact scenario the comment at `:44-50` exists to prevent.

Fixing #4 (`{ expire: 0 }`) closes the same-process window. The remaining ask is that this read must not be allowed to silently return defaults — see the note in *What I'd do next*.

### 6. Four copies of the default `sectionOrder`, one of which disagrees

`lib/site-features.ts:34-43` is `search, collection, lightbox, news, blogs, featureArticles, events, about`. `config/site-features.json:12-21`, api#149's `DEFAULT_SITE_FEATURES` (`apps/common/views.py:239-248`) and api#149's seed migration (`0010_seed_site_features.py:34-43`) are all `search, lightbox, collection, blogs, featureArticles, about, news, events`. Only `ALL_SECTION_KEYS` is the odd one out, and `normalizeSectionOrder(undefined)` appends in that order too — so it is what every degraded read serves. One-array fix; `lib/site-features.test.ts` asserts relative to `ALL_SECTION_KEYS` (`:22,26,73,112`) rather than a literal, so it stays green.

### 7. Backend 4xx on the write is flattened to a 502 and the reason is thrown away

`app/api/app-settings/route.ts:62-64` binds nothing and logs nothing, discarding the `Failed to write site features: ${res.status}` from `lib/site-features-server.ts:119-121`. The try/catch itself is new and an improvement — before this PR an fs write failure just 500'd — but api#149 raises `ValidationError` (400) on a bad payload shape and DRF yields 403 if `IsSuperuserOrReadOnly` disagrees with the frontend's profile check, and both reach the admin as the same opaque toast (`page.tsx:44-47,83-85`). In dev the server log is no help either: `apiFetch`'s error branch is unreachable when `isDev`, so the only line printed is `[API] OK PUT /api/v1/app-settings/ → 400`. At minimum `catch (err) { console.error('[site-features] backend write failed', err); … }`.

### 8. `res.json()` rejecting still degrades the whole site with clean logs

`lib/site-features-server.ts:87-89`. The new log at `:49-54` covers a *parsed* non-object body. A body that fails to parse at all — nginx truncating a response, a worker killed mid-stream, an HTML error page — rejects at `:45` and lands in the unbound `catch`. `apiFetch` can't cover it; its catch (`lib/api-fetch.ts:43-49`) wraps only the `fetch()` call and it has already returned the Response.

<details>
<summary><b>Nine smaller things</b></summary>

- **`lib/site-features-server.ts:102-113`** — the "strict whitelist" only constrains the four top-level keys. `sections` (`:103`) and `searchCategories` (`:112`) are forwarded verbatim, and the route's shape guard is just `'sections' in body && 'searchCategories' in body` (`route.ts:41`). api#149's `flatten_settings` (`apps/common/views.py:182-196`) recurses into whatever nesting arrives, and `put` then deletes every `site_features.*` row not in the flattened result — so a payload like `searchCategories: {manuscripts: {enabled: {a: {b: 1}}}}` from a script or a half-hydrated form wipes every visibleColumns/visibleFacets row. The durable fix is in #149's serializer, not here — but the comment at `:97-101` overstates what the whitelist does.
- **`lib/api-fetch.ts:37-41,47`** — `apiFetch` is isomorphic (`API_BASE_URL` at `:19` branches on `typeof window`), and nine client components route through it: `app/(site)/hands/[id]/hand-viewer.tsx` via `apiFetch`, plus eight backoffice pages/components via `authFetch`. The new non-2xx `console.error` and the now-unconditional catch log therefore also run in the browser, where they do nothing for the server-log visibility this was written for. The public-site case is `hand-viewer.tsx:109,112`: a visitor whose graph request 500s gets `[API] GET /api/v1/manuscripts/graphs/?hand=… → 500` in their own console while the server log this was written for records nothing. Gate the two production branches on `typeof window === 'undefined'`. (`contexts/search-context.tsx`, `components/manuscript/viewer-text-panel.tsx`, `components/backoffice/texts-list.tsx` and `hooks/search/use-search-export.ts` import only `API_BASE_URL` and use raw `fetch`, so they're unaffected — and get none of this logging either.)
- **`lib/api-fetch.ts:34-36`** — the dev tag derives purely from duration, so a fast 500 prints `[API] OK GET … → 500 (12.4ms)` at `console.log` level, and the new error branch is unreachable in exactly the environment where someone is watching. `const tag = !res.ok ? 'ERR' : duration > SLOW_THRESHOLD ? 'SLOW' : 'OK'`.
- **`lib/api-fetch.ts:7-11`** — the rationale comment names `readModelLabels()`, which never goes through `apiFetch`: `lib/model-labels-server.ts:1,10-23` reads `config/model-labels.json` via `readJsonConfig`. Someone chasing a silent model-labels degradation will read this, expect a log, find none, and conclude the read succeeded. (`readSiteFeatures()` and `getPublishedPages()` in the same sentence are both correct — `lib/pages-server.ts:14`.)
- **`app/api/site-features` → `app/api/app-settings` with no alias** — nothing remains at the old path (`git ls-tree -r pr/111 -- app/api`) and `next.config.mjs` adds no rewrite. During a rolling deploy an already-open backoffice tab PUTs the old path, gets the 404 HTML page, `res.json()` rejects, and `page.tsx:44-47` surfaces *"Unknown error"* with no hint that a reload fixes it. The noun also now disagrees with `SITE_FEATURES_PATH`, `SITE_FEATURES_TAG`, `lib/site-features-server.ts` and `/backoffice/site-features` — worth picking one.
- **`config/site-features.test.ts:44-48`** — after this PR nothing reads `config/site-features.json` (`git grep` at this head returns only `.prettierignore:24`, this test, and `docs/SEARCH_FEATURE_CONFIG_OWNERSHIP.md:11,17`). The "seeds every manuscripts facet as visible to anonymous visitors" case is now a CI guard with zero runtime effect: add a facet, this test fails, you add it to the JSON, CI goes green — and anonymous visitors still can't see it, because the live value is an AppSettings row. That's the exact silent breakage the test was written to catch.
- **`lib/site-features-server.ts:1`** — the module lost its implicit server-only guard. It used to import `./json-config-file` → `fs/promises`, so an accidental client import failed the build; it now imports the isomorphic `./api-fetch`, so the same mistake compiles and ships a cross-origin browser fetch on every mount.
- **`app/api/app-settings/route.ts:7-16`** — `verifySuperuser` is now byte-identical in four route handlers (`app-settings`, `model-labels`, `pages`, `pages/[slug]`). Before this PR the site-features copy at least differed (`is_staff` at `2e1fb3d:app/api/site-features/route.ts:12`), so the duplication was visible.
- **`lib/site-features-server.ts:122-126`** — the comment says callers can "echo back exactly what was persisted", but the function never reads `res.json()`; it returns its locally-derived `normalized`. api#149's `put` re-reads the rows and re-runs `unflatten_settings` for its response, and `page.tsx:79` seeds `queryClient.setQueryData` with the frontend's view instead — which is the cache divergence the comment claims to prevent. Docs are stale too: `docs/SEARCH_FEATURE_CONFIG_OWNERSHIP.md:11,17` still calls the JSON a runtime override layer, `docs/TECHNICAL_ISSUES_AND_REMEDIATION.md:93,97,272` still names `app/api/site-features/route.ts` (and `:47,213` the old URL), and the PR body says the backend adds `/api/v1/site-features/` when both sides use `/api/v1/app-settings/`.

</details>

---

## What I'd do next

1. Hold the merge until api#149 lands, or merge them together.
2. Split `readSiteFeatures` into a strict `fetchSiteFeatures()` that throws and a `readSiteFeatures()` that catches. Public renders keep the lenient one; the GET route and the PUT's pre-write read use the strict one and answer 503, and the backoffice renders a load error instead of an editable form when the query fails. That kills #2 and #5 at the root.
3. Add a module-scope last-known-good so a blip holds the previous state instead of unlocking everything (#3) — three lines. It's per-process and lost on restart, so it bounds the fail-open window rather than closing it; worth having, not worth describing as a guarantee.
4. `revalidateTag(SITE_FEATURES_TAG, { expire: 0 })` and trim the comment (#4). Keep `revalidatePath('/', 'layout')` — it's load-bearing today, including for the PUT's own pre-write read.
5. Reorder `ALL_SECTION_KEYS` to match the other three copies (#6).

Happy to push 2-5 plus tests (the 503 paths, the last-known-good behaviour, and the `revalidateTag` argument — the existing `next/cache` mock is a bare `vi.fn()` that is never asserted, which is why the wrong profile went unnoticed) onto `fix/wire-site-features-to-backend2` if that's less friction than doing it yourself. Say the word.

PR #117, opened 2026-08-17
feat(publications): display keywords and themes on publication cards and detail pages

### LINKED ISSUE #80: Publications: themes/keywords are missing (Blog, News, Feature Articles)
**News/Blog/Feature Articles.** @pastokes: *"The themes/keywords are missing (same for Blogs and News)."* Surface themes/keywords on publications (may also require the field to be imported/exposed by the API).

---
From [*UI feedback* · project-discussions#8](https://github.com/orgs/archetype-pal/discussions/8), reported by @pastokes (2026-06-29).

lanterno:
Closes #80

### Description
Surfaces publication keywords/themes as subtle tag badges with a tag icon on publication previews (Blog, News, Feature Articles) and detail pages.

### Changes
- `BlogPostPreview`: Added `keywords?: string | null` prop and tag rendering with `Badge` component and lucide `Tag` icon.
- `BlogPostPreview`: Added keyword string parser supporting comma separation and quotes.
- `PublicationDetailPage`: Passed `keywords={item.keywords}` to `BlogPostPreview`.
- `PaginatedPublications`: Passed `keywords={article.keywords}` to publication card `BlogPostPreview`.
- Added translations `content.blog.keywordsLabel` for English (`Keywords`) and French (`Mots-clés`).
- Unit tests added in `components/content/blog-post-preview.test.tsx`.

### DIFF
--- a/components/content/blog-post-preview.test.tsx
+++ b/components/content/blog-post-preview.test.tsx
@@ -23,4 +23,76 @@ describe('BlogPostPreview', () => {
     expect(html).toContain('href="#refnote1"');
     expect(html).toContain('href="/manuscripts/259"');
   });
+
+  it('renders keywords as badges when keywords string is provided', () => {
+    const html = renderToStaticMarkup(
+      <BlogPostPreview
+        title="Medieval Manuscripts"
+        author="Alice Smith"
+        date="2020-01-01T00:00:00Z"
+        excerpt="<p>An overview.</p>"
+        slug="/publications/blogs/medieval-manuscripts"
+        keywords="palaeography, charters, medieval history"
+        showShareBtns={false}
+        showReadMoreBtn={false}
+      />
+    );
+
+    expect(html).toContain('data-testid="publication-keywords"');
+    expect(html).toContain('palaeography');
+    expect(html).toContain('charters');
+    expect(html).toContain('medieval history');
+  });
+
+  it('handles quotes and whitespace in keywords gracefully', () => {
+    const html = renderToStaticMarkup(
+      <BlogPostPreview
+        title="Medieval Manuscripts"
+        author="Alice Smith"
+        date="2020-01-01T00:00:00Z"
+        excerpt="<p>An overview.</p>"
+        slug="/publications/blogs/medieval-manuscripts"
+        keywords='"insular script", "anglo-saxon", latin'
+        showShareBtns={false}
+        showReadMoreBtn={false}
+      />
+    );
+
+    expect(html).toContain('insular script');
+    expect(html).toContain('anglo-saxon');
+    expect(html).toContain('latin');
+    expect(html).not.toContain('&quot;insular script&quot;');
+  });
+
+  it('does not render keywords container when keywords is undefined or empty', () => {
+    const htmlEmpty = renderToStaticMarkup(
+      <BlogPostPreview
+        title="No Keywords"
+        author="Bob"
+        date="2020-01-01T00:00:00Z"
+        excerpt="<p>None.</p>"
+        slug="/publications/blogs/no-keywords"
+        keywords=""
+        showShareBtns={false}
+        showReadMoreBtn={false}
+      />
+    );
+
+    expect(htmlEmpty).not.toContain('data-testid="publication-keywords"');
+
+    const htmlNull = renderToStaticMarkup(
+      <BlogPostPreview
+        title="No Keywords"
+        author="Bob"
+        date="2020-01-01T00:00:00Z"
+        excerpt="<p>None.</p>"
+        slug="/publications/blogs/no-keywords"
+        keywords={null}
+        showShareBtns={false}
+        showReadMoreBtn={false}
+      />
+    );
+
+    expect(htmlNull).not.toContain('data-testid="publication-keywords"');
+  });
 });
--- a/components/content/blog-post-preview.test.tsx
+++ b/components/content/blog-post-preview.test.tsx
@@ -1,8 +1,34 @@
 import * as React from 'react';
+import { render } from '@testing-library/react';
 import { renderToStaticMarkup } from 'react-dom/server';
 import { describe, expect, it } from 'vitest';
 
 import BlogPostPreview from './blog-post-preview';
+
+const renderPreview = (keywords?: string | null) =>
+  render(
+    <BlogPostPreview
+      title="Medieval Manuscripts"
+      author="Alice Smith"
+      date="2020-01-01T00:00:00Z"
+      excerpt="<p>An overview.</p>"
+      slug="/publications/blogs/medieval-manuscripts"
+      keywords={keywords}
+      showShareBtns={false}
+      showReadMoreBtn={false}
+    />
+  );
+
+/** The badge labels, in order, or null when the group is not rendered at all. */
+const badges = (keywords?: string | null): string[] | null => {
+  const { container, unmount } = renderPreview(keywords);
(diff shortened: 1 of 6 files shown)


lanterno (commented):
Both call sites are covered (`paginated-publications.tsx`, `publication-pages.tsx` — those are the only two `BlogPostPreview` usages), the field is already on `Publication` in `utils/api.ts` and returned by both the list and detail serializers, so the wiring is complete and CI is green. The placement between the meta row and the body reads well. Four things before merge.

---

### 1. The parser doesn't match the format the API actually emits (correctness)

`parseKeywords` (`components/content/blog-post-preview.tsx:23`) splits on commas first, then strips one leading/trailing quote. But `keywords` comes from Tagulous' `TagStringField.to_representation` → `str(value)` → `render_tags`, whose contract is (`tagulous/utils.py:218`):

- tags joined with `", "`, sorted;
- **any tag containing a comma _or_ a space is wrapped in `"…"`**;
- inner quotes escaped by doubling (`"` → `""`).

Splitting on comma before honouring the quotes breaks two of those three rules. I ran the PR's parser against `render_tags` output:

| DB tags | API string | PR renders |
|---|---|---|
| `insular script`, `latin` | `"insular script", latin` | ✅ `insular script`, `latin` |
| `edinburgh, scotland`, `latin` | `"edinburgh, scotland", latin` | ❌ `edinburgh`, `scotland`, `latin` — **one tag becomes two** |
| `great "seal"` | `"great ""seal"""` | ❌ `great ""seal""` — doubled quotes leak into the UI |

Not hypothetical: publication 33 in the migrated corpus carries a tag literally named `,`, so the API sends `",", 1, 2, a, g, t` and the tag is silently dropped (the two halves strip to empty). A neighbouring space-named tag survives `filter(Boolean)` as an invisible badge, because the trim happens *before* the quotes are stripped.

A scanner that mirrors `render_tags` is about the same size and round-trips all of the above (including both of this PR's own test inputs):

```ts
const parseKeywords = (value?: string | null): string[] => {
  if (!value) return [];
  // Mirrors Tagulous' render_tags on the API side: tags join with ", ", any tag
  // containing a comma or a space is quoted, and inner quotes are doubled.
  const tags: string[] = [];
  let tag = '';
  let inQuote = false;
  for (let i = 0; i < value.length; i++) {
    const c = value[i];
    if (c === '"') {
      if (value[i + 1] === '"') {
        tag += '"';
        i++;
      } else {
        inQuote = !inQuote;
      }
    } else if (c === ',' && !inQuote) {
      tags.push(tag);
      tag = '';
    } else {
      tag += c;
    }
  }
  tags.push(tag);
  return tags.map((t) => t.trim()).filter(Boolean);
};
```

(It also stops stripping `'`, which Tagulous never emits as a delimiter — the current regex corrupts a tag legitimately named `'foo'`.)

### 2. On the real corpus this mostly renders the publication type as a keyword

Counted in the migrated MoA data: **53 of the 58 publications that have keywords have _only_ `News`, `Blog`, or `Feature Article`** — the exact strings already shown, three items to the left, as `publicationLabel` in the meta row. So for the overwhelming majority of pages this ships a badge that repeats the type, and the handful of genuinely thematic tags (`palaeography` ×4, `scribes` ×3, `medieval` ×1) plus publication 33's junk (`t`, `a`, `g`, `1`, `2`) are the rest.

That's the data half of #80 — the issue itself flags it: *"may also require the field to be imported/exposed by the API"*. Exposing it is done here; importing the real themes isn't. Two options, and I'd take both:

- filter the three category names out of the badge row (they're duplicated by the type label either way), so the feature degrades to "no badges" rather than "a badge that says News";
- open a data-migration issue for the actual themes, and link it from #80 so the reporter's ask isn't closed by the display half alone.

Worth noting I could only check the dev corpus — production was returning 502 on every path while I was reviewing, so I couldn't confirm the same distribution there.

### 3. `content.blog.keywordsLabel` is added but never used

Added to `messages/en.json` and `messages/fr.json` (and listed in the PR description as a change), but there is no `keywordsLabel` reference anywhere in the codebase — the badge row renders with no label, visible or otherwise. Right now it's an orphan key in both catalogues.

The two problems solve each other: the group is a bare run of words after the byline with nothing naming it for a screen reader, so put the key on the container.

```tsx
<div
  className="flex flex-wrap items-center gap-1.5 mb-4"
  aria-label={t('blog.keywordsLabel')}
  data-testid="publication-keywords"
>
```

With a group label, the `Tag` icon repeated on *every* badge (`:98`) also stops earning its place — one leading icon, or none, would be quieter. Non-blocking design call, your judgement.

### 4. No test covers the format the API actually sends

The three new cases use hand-typed strings (`palaeography, charters, medieval history`, `"insular script", "anglo-saxon", latin`) — both are shapes a human types into the backoffice field, and both already pass. Nothing exercises a quoted tag containing a comma, which is exactly where finding 1 bites. Worth one case pinned to `render_tags` output:

```tsx
keywords='"edinburgh, scotland", latin'   // → two badges, not three
```

Also `expect(html).toContain('charters')` matches anywhere in the document; asserting against the keywords container (the `data-testid` is already there for it) would make these guards rather than smoke tests.

---

1 and 3 I'd fix before merge; 2 is the one that decides whether the feature actually answers @pastokes; 4 follows from 1.

lanterno:
Addressed in 7c158a1, on this branch.

| # | Finding | Status |
|---|---|---|
| 1 | Parser doesn't match Tagulous' `render_tags` format | **Fixed** — replaced with a scanner that honours quoting and `""` escapes |
| 2 | 53/58 publications badge only their own category | **Fixed (display)** + data issue opened |
| 3 | `content.blog.keywordsLabel` orphaned in en+fr | **Fixed** — now the group's `aria-label` |
| 4 | No test for the format the API sends | **Fixed** — 6 cases, scoped to the container |

**1.** `parseKeywords` now walks the string instead of splitting it: a comma inside quotes doesn't delimit, and `""` decodes to a literal `"`. Round-trips every case I raised — `"edinburgh, scotland", latin` → two badges, `"great ""seal"""` → `great "seal"`. It also stops stripping `'`, which Tagulous never emits as a delimiter.

**2.** The badge row now drops `news` / `blog` / `feature article` (case-insensitively) — the meta row already shows the type three items to the left, so badging it again was pure duplication. This is a display-layer stopgap and the comment says so; the real fix is importing the actual themes, which I've split out as [archetype-data-migration#5](https://github.com/archetype-pal/archetype-data-migration/issues/5) and linked from #80. Net effect on today's data: the 53 publications that only carried a category name now show no badges instead of a badge reading "News". If you'd rather ship the duplication until the import lands, it's one line to drop.

**3.** `aria-label={t('blog.keywordsLabel')}` on the container — the key you added is now used, and the badge run has a name instead of being a bare list of words after the byline.

**4.** The keyword tests moved from `renderToStaticMarkup` + whole-document `toContain` to Testing Library, asserting the badge labels in order inside `[data-testid="publication-keywords"]`, so they fail on a wrong split rather than passing because the word appears in the title. Covers the quoted-comma form, doubled quotes, category filtering, absence for empty/null/undefined, and the `aria-label`. Three of them fail against 37b923b and pass on 7c158a1 — checked by stashing the component.

**Left alone:** the `Tag` icon on every badge. I flagged it as a design call and it's still yours to make — with a group label it's arguably redundant, but that's a look-at-it-in-the-page judgement, not something to decide from a diff.

Gates run locally before pushing: `tsc --noEmit` clean, eslint clean, prettier clean, full suite 125 files / 1256 tests green, en/fr parity 2353 each with no orphaned keys.

PR #118, opened 2026-08-17
feat(search): add clause grid cards with region cropping and thumbnail toggle (frontend#74)

### LINKED ISSUE #74: Clauses grid: show the region (not the full image) + restore 'show without thumbnail'
**Clauses.** @pastokes:
- *"Grid view thumbnail images are not working (they show the full image, not the region), but I assume this is because the text regions are not imported yet."* — thumbnail should be cropped to the clause's text region (likely blocked on text-region import data).
- *"There was an option to show the clause without the thumbnail image, but this no longer seems available."* — restore the text-only (no thumbnail) view option.

---
From [*UI feedback* · project-discussions#8](https://github.com/orgs/archetype-pal/discussions/8), reported by @pastokes (2026-06-29).

lanterno:
Closes archetype-pal/frontend#74

### Summary of Changes

1. **Clause Grid Cards & Region Cropping**:
   - Added `ClauseListItem` support to `SearchGrid`, mapping clause search results into `ClauseGridCard`.
   - Used `useIiifThumbnailUrl(clause.thumbnail_iiif, clause.annotation_coordinates)` so clause thumbnails are cropped to the text region coordinates when present.
   - Displayed clause header label (`clause_type`, `shelfmark`, `locus`), highlighted text content snippet, and collection / lightbox triggers for connected annotations.

2. **Restore "Show without thumbnail" (Text-only) View Option**:
   - Added `useShowThumbnails` hook using `useSyncExternalStore` persisted to `localStorage` to support toggling between image thumbnails and text-only mode across search results.
   - Added `ThumbnailToggle` button to the search header toolbar with i18n support in English and French (`messages/en.json` and `messages/fr.json`).
   - In `ResultsTable`, suppressed the image preview row when `showThumbnails` is false so tabular view shows clean text without image previews.
   - In `SearchGrid`, rendered cards without image containers when `showThumbnails` is false.

3. **Testing & Validation**:
   - Added unit tests for clause grid card mapping and highlight formatting in `search-grid-label.test.ts`.
   - Added unit tests for `ThumbnailToggle` component in `thumbnail-toggle.test.tsx`.
   - Added unit tests for `useShowThumbnails` in `use-show-thumbnails.test.ts`.
   - Verified 127/127 test files (1,258 tests) pass, ESLint 0 errors / 0 warnings, and Next.js `pnpm build` succeeds.


lanterno (commented):
Reviewed at `1d51331`. Gates re-run locally: `tsc --noEmit` clean, `eslint .` 0/0, `prettier --check` clean, search tests 8 files / 44 tests green, en/fr parity 2358 keys each.

The clause half is the right fix and worth calling out: `toGridCard` previously fell through to `return null` for `clauses`, so the clause grid rendered *nothing* — `flatCards` filtered every card out. That was the "grid thumbnails not working" report, and cropping through `useIiifThumbnailUrl(thumbnail_iiif, annotation_coordinates)` matches how `AnnotationInlinePreview` already crops in table view. Good.

The thumbnail-toggle half needs work before merge. Six findings, roughly in severity order.

---

### 1. Text-only mode silently removes "add to collection" and "open in lightbox" (functional regression)

In `MediaGridCard` the action overlay is *inside* the thumbnail block (`components/search/search-grid.tsx:238`), and `ClauseGridCard` repeats the same structure (`:424`). Turning thumbnails off unmounts the `OpenLightboxButton` and `CollectionStar` along with the image.

Table view has the same hole from the other direction: `results-table.tsx:594` now gates the whole preview row on `showThumbnails`, and for `clauses` the connected-graph star at `:611` lives *only* in that row. (`images` and `graphs` keep their stars — those are in `COLUMNS`, at `:130` and `:212` — so the loss is specific to clauses.)

Net effect: a user who prefers the compact text list can no longer star a clause or a graph, or send one to the lightbox, from either view. The control presents itself as a display-density preference, so losing a primary feature behind it will read as a bug. The actions want to move out of the image container — right-aligned in the card footer when the thumbnail is hidden, and kept next to the row when the table preview is hidden.

### 2. Text-only mode still fetches every IIIF `info.json` (performance)

`GraphGridCard:289` and `ClauseGridCard:397` call `useIiifThumbnailUrl(...)` unconditionally, before anything branches on `showThumbnail`. When `coordinates` are present `getIiifImageUrlWithBounds` **awaits `fetchIiifImageInfo`** (`utils/iiif.ts:272`) — a real network round-trip per distinct image. `IIIF_INFO_CACHE` dedupes repeats but not first loads, so a 50-row clause page across 50 images issues 50 requests whose results are then thrown away.

Avoiding that work is the entire point of the option the reporter asked for. The hook already short-circuits on an empty `infoUrl` and returns `null`, so `useIiifThumbnailUrl(showThumbnail ? infoUrl : '', …)` is the whole fix.

### 3. The toggle renders in views where clicking it does nothing (UX)

`ThumbnailSizeControl` is carefully gated to `viewMode === 'grid'` (`search-page.tsx:249`); `ThumbnailToggle` at `:256` is gated on nothing. It therefore shows in `timeline`, `distribution`, and `map`, where no thumbnail exists to hide.

Table view is only partly better: the preview row exists only for the four types with a `previewAccessor` — `texts`, `clauses`, `people`, `places` (`results-table.tsx:374-417`). For `images`, `manuscripts`, `graphs`, `scribes`, and `hands` in table view the button is equally inert.

Worth gating on grid-view or a table view that actually has a preview.

### 4. It is a toggle, but doesn't announce itself as one — and its accessible name contradicts its visible label (a11y)

`thumbnail-toggle.tsx` has no `aria-pressed`. The codebase has the convention both ways already (`ThumbnailSizeControl` uses `role="radio"`/`aria-checked`; `app/(site)/collection/page.tsx:773` uses `aria-pressed`), so a screen-reader user currently gets a plain button with no state.

Worse, `aria-label` overrides the visible text. With thumbnails off the button *reads* "Text only" but its accessible name is "Show thumbnails" — no word in common. That's a WCAG 2.5.3 (Label in Name) failure: a voice-control user saying "click Text only" gets no match.

Dropping `aria-label` entirely is the fix — the visible text already is the label. Keep `title` as the hint, add `aria-pressed={showThumbnails}`. That also makes the `showThumbnails`/`hideThumbnails` message keys redundant.

### 5. Nothing tests the behaviour that can regress

The three new tests cover the hook's persistence and the button's own label. Neither `SearchGrid` nor `ResultsTable` is ever rendered with `showThumbnails={false}` — which is the only place findings 1 and 2 could have been caught. `search-grid-label.test.ts` exercises `toGridCard`, a pure mapping function; `ClauseGridCard` itself never renders in a test.

### 6. A third localStorage idiom, and a default that isn't sticky (nit)

There are now three near-identical persisted-preference stores with three shapes: `useThumbnailSize` (useState + effect, key `search-grid-thumbnail-size`), the sidebar flag in `use-search-view-mode.ts` (module-level `useSyncExternalStore`, key `search-view-prefs`), and this one (a copy of the second, key `search-show-thumbnails`).

Two small things inside the new one:

- `getSnapshot` returns `true` only for the literal string `'true'`, so any corrupted value lands on text-only rather than on the documented default of thumbnails-on. `saved !== 'false'` is the safer read.
- On a failed write (Safari private mode, storage disabled) `change` still notifies, `getSnapshot` re-reads the old value, and the button appears dead. `useThumbnailSize` degrades gracefully here because it holds React state; this one doesn't.

### Not blocking, author's call

`className="hidden shrink-0 sm:inline-flex"` (`search-page.tsx:258`) hides the toggle below `sm`. That was copied from the size control, but the rationale doesn't transfer — a text-only list is arguably *most* useful on a narrow screen or a slow connection, which is close to the original DigiPal motivation. Left alone since it's a layout judgement about a crowded header row.

lanterno:
Pushed `f44b5b9` addressing the review.

| # | Finding | Status |
|---|---------|--------|
| 1 | Text-only mode dropped collect/lightbox | Fixed — actions moved to the card footer; the table's preview row now survives as just the star whenever there's something to collect |
| 2 | Text-only mode still fetched every `info.json` | Fixed — `useIiifThumbnailUrl(showThumbnail ? infoUrl : '', …)` in both `GraphGridCard` and `ClauseGridCard` |
| 3 | Toggle rendered where it does nothing | Fixed — gated on grid view or a table type with a preview row, via a new exported `hasTablePreview()` |
| 4 | No `aria-pressed`; accessible name contradicted the visible label | Fixed — `aria-pressed` added, `aria-label` dropped, and the two message keys that existed only for it removed (en/fr parity 2356 each) |
| 5 | Nothing tested the behaviour that can regress | Fixed — `search-grid.test.tsx` and `results-table-preview.test.tsx`; verified 3 of the 4 grid assertions fail against `1d51331` and pass after |
| 6 | Non-sticky default | Fixed the narrow half — only an explicit `'false'` turns thumbnails off. Left the three-localStorage-idioms consolidation and the silent-failed-write degradation alone: both are pre-existing patterns and unifying them is a wider refactor than this PR's scope |

Not changed, still your call: the toggle stays `hidden … sm:inline-flex`, so it's unavailable below `sm`. Making it visible on mobile is where a text-only list arguably earns the most, but it's a judgement about a header row that's already crowded, and it wants an eye on the actual layout rather than a blind class change.

Gates on `f44b5b9`: `tsc --noEmit` clean, `eslint .` 0 errors / 0 warnings, `prettier --check` clean, full suite 129 files / 1266 tests green.

PR #119, opened 2026-08-17
Add German (de) as a supported UI language

geourjoa:
## Summary
- Adds `de` to the `Locale` type/list (`lib/locale.ts`), which drives next-intl's request config, the message-parity test, and the locale cookie store.
- Adds `messages/de.json` — a full German translation mirroring `en.json`'s exact key structure (validated against the existing `lib/i18n-messages.test.ts` parity test).
- Replaces the hardcoded EN/FR toggle button in `language-switcher.tsx` with a dropdown listing all supported locales.
- Extends the trilingual content model (site labels, CMS pages) with a German field: `lib/model-labels.ts`, `config/model-labels.json`, `lib/pages.ts`, and their backoffice editors (translations page, `pages/[slug]` editor, `pages/new`).
- Companion backend PR backfills German site-label values: archetype-pal/backend https://github.com/archetype-pal/backend/pull/179


geourjoa:
Rebased onto current `main`. The branch had originally forked from a local checkout that included unmerged, unpushed WIP commits (trash page, impersonation UI details, etc.) alongside this change — that's now fixed; the branch is a single commit sitting directly on `origin/main`, and `messages/de.json` has been re-synced to match `main`'s actual `en.json` key set (2 keys added, ~55 not-yet-upstream keys removed). Diff is now 16 files / +2847 -67, down from the earlier ~7000-line diff.

lanterno (commented):
Reviewed the full diff and re-ran the gates against a **live merge with current `main`**, because the green checks on this PR are from 2026-08-18 and the branch is 26 commits behind. That re-run is where the one blocking item shows up.

---

## Blocking: `messages/de.json` is 61 keys behind `main`

`mergeStateStatus` is `CLEAN`, but that only means there is no *textual* conflict. The conflict here is semantic and git cannot see it: `main` has added 61 new message keys to `en.json`/`fr.json` since this branch was cut, and `de.json` has none of them. It also still carries one key `main` has since renamed.

Merging `origin/main` into `worktree-add-german-language` locally and running the suite gives:

```
Test Files  1 failed | 132 passed (133)
     Tests  1 failed | 1297 passed (1298)

FAIL lib/i18n-messages.test.ts > locale "de" has the same key structure as the default locale
```

```
en: 2413   fr: 2413 (missing 0, extra 0)   de: 2353 (missing 61, extra 1)
```

The missing keys are the ones added by the trash/annotations work, the thumbnail toggle and the publication-keywords card:

- `backoffice.trash.*` — 38 keys (the entire Trash page: title, filters, columns, restore/purge, all toasts)
- `manuscript.delete.*` / `manuscript.gallery.*` — 14 keys (delete confirmations, region trash toasts)
- `backoffice.annotations.toastMovedToTrash`, `bulkTrashed`, `bulkTrashFailed`, `bulkPartial`
- `search.showThumbnailsTitle`, `hideThumbnailsTitle`, `thumbnailsOn`, `textOnly`
- `content.blog.keywordsLabel`
- **extra:** `backoffice.annotations.toastDeleted` — renamed to `toastMovedToTrash` on `main`, so this one is now dead weight

Two consequences, not one. The parity test failing turns `main` red on merge, and that test exists precisely because of the second: with next-intl's defaults (`defaultGetMessageFallback` returns the dotted path, `defaultOnError` just `console.error`s — verified in `use-intl@4.13.0/dist/esm/development/initializeConfig-*.js`), a German visitor does not get an English fallback and does not get a crash. They get the literal string `backoffice.trash.title` rendered in the UI, for all 61.

The fix is mechanical — translate the 61, drop the 1 — but it needs redoing on whatever day this actually merges, since `main` keeps moving.

---

## Non-blocking

**1. `footerBottomLine` default was rewritten, and it is not in the description.** `lib/model-labels.ts` replaces the shipped English/French defaults — `©2015–17 Models of Authority. Some parts available under CC-BY licence. All manuscript images are copyright of their respective repositories. Website by DDH / KDL. Built with Archetype.` — with a bare `Built with Archetype.` / `Construit avec Archetype.`

I checked the impact rather than guessing: `common_sitelabel` row 22 stores the full text for `en`/`fr`, and `resolveModelLabel` reads the stored value ahead of the default, so **no visible change for English or French visitors on this deployment**. As a platform default it is arguably the better value — hardcoding one instance's 2015–17 copyright as the shipped default of a reusable product is a bug in its own right. But it is a deliberate content change to a legal/attribution string sitting inside an i18n PR, and it only stays invisible while the DB override exists. It belongs in the description, or in its own commit.

Worth knowing where it does surface: German has no stored value yet, so `de` resolves to the new default. With backend #179 as written, German visitors get `Fußzeile, unterer Abschnitt` in that slot — a placeholder, not the attribution. That is a #179 data question, not a defect here, but the two PRs land together.

**2. `messages/en.json` — `login.signIn` was `"Se connecter"`, now `"Sign in"`.** A real bug fix; French had leaked into the English catalogue. Also unmentioned in the description. Good catch, just flag it next time — a one-line copy fix is easy to lose inside +2897.

**3. The description lists `config/model-labels.json` as extended.** That file does not exist in this repo and is not in the diff. Harmless, but the description should match.

**4. `lib/model-labels-server.test.ts:38`** swapped a literal assertion for `expect(config.labels.appManuscripts).toEqual(DEFAULT_MODEL_LABELS.appManuscripts)`. That now compares the code against itself, so it can no longer catch an unintended edit to the default. The literal was the stronger assertion.

**5. `language-switcher.tsx` was rewritten** from a two-state toggle to a dropdown and still has no test — it had none before, so this is not a regression, just the moment where one became worth having. Losing `title` as an action hint (`"Passer en français"` → the current language's own name) is a small a11y downgrade for a control whose visible label is a bare `DE`.

---

## What I verified

- **de.json quality is genuinely good.** No empty values, no non-strings, and of 142 keys identical to English only 4 are ≥3 words — `RSS / Atom`, `Status → {status}`, `#{id} → {target}`, `la, en, enm…` — all correctly locale-neutral. For comparison `fr` has 187 identical. This is a real translation, not a copy of `en`.
- **ICU placeholders and rich-text tags match `en` on every one of the 2353 keys**, in both `fr` and `de`. This matters because the parity test only compares key *paths* — a `{count}` renamed inside a German string would pass that test and throw at render. None do.
- **Locale plumbing is safe.** `lib/i18n.ts` and `stores/locale-store.ts` both validate the cookie against `locales` and fall back to `defaultLocale`, so a stale or hand-edited `NEXT_LOCALE` cannot reach a missing catalogue.
- **Runtime back-compat holds** for the widened `LocalizedLabel`/`LocalizedText`: `normalizeLocalizedValue` and `normalizeLocalizedText` both default `de` when the backend omits it, which is exactly what the current DB returns.
- **Gates on the merged tree:** eslint clean, prettier clean, `tsc --noEmit` produces only two pre-existing `CarouselItem.weight` errors that are already on `main` (unrelated to this PR — worth a separate issue, since `next build` does not typecheck test files and CI therefore misses them). Tests: 1297/1298, the single failure being the parity test above.

Net: the PR itself is well-built and the translation is careful work. The only thing standing between it and a merge is 61 keys of drift that accumulated while it waited.

geourjoa:
@lanterno German missing translation are fixed, ready to merge !

lanterno:
Yes, just did! :)

Issue #110, opened 2026-08-10
Allow user to compare image side by side


geourjoa:
As discussed in August 10th meeting.

We want to be able to compare image side by side. We imagine two ways to do that : 
- Use current OpenSeaDragon (OSD)
- Use Mirador viewer

We only talk about viewing or comparing workflow. Annotating is out of scope 

Mirador
-  pros : 
    - Seems easy to implement, almost embed it and pass the right config (the list of wanted images basically)
   - Mirador is a dedicated tool to view IIIF media so we will benefit all of the functionality we want 
- cons 
    - New dependancy in the stack 
    - Will need to write an annotation adapter ([dumb example](https://github.com/TETRAS-IIIF/mirador-annotation-editor/blob/main/src/annotationAdapter/LocalStorageAdapter.js) )
 
OSD 
- pros :  
    - Already in the project 
    - More customizable
- cons
    - We will need to write a Mirador-like viewer

geourjoa:
Thanks @sjabrookes for the details

We need to precise my point 2. Manifest usage or use direct IIIF image link and build around 

I dig the code and I'm almost sure Archetype use IIIF image but not handle IIIF manifest in the current state. I dont know how much you are familiar with IIIF manifest. To summarize, it's a json file composed with at the least one canvas (a canvases is a media with a height and width or duration). Canvas can have annotations with different `motivation` (the viewer will handle it differently dependent on that) and target (spatial xywh or temporal)

In archetype the object representing the data of a manifest is a manuscript. The "IIIF canvas" equivalent are represented by `manuscript_item parts`. For the annotation, its `annotation_graph` but there are not in IIIF format. On archetype, these data are reprensented in a relational table.

So right now we have no way to use Mirador since Archetype has no IIIF manifest. Mirador need to have a IIIF manifest URL as entry to work. 

If we want to use Mirador we need to produce and expose IIIF manifests 

--- 

We can discuss about that today

geourjoa:
@sjabrookes 

I find very exciting thing in the backend : 

The manifest.py is almost exactly what i wanted. The endpoint expose the manifest following IIIF presentation 3 API

What to call on the backend : 

https://digipal-api.tetras-libre.fr/api/v1/iiif/item-parts/1/manifest

My mirador to test : 
https://tetras-iiif.github.io/mirador-video/latest

A bit surprisingly I dont find any use of this code in the frontend  so far. Was done by @lanterno

geourjoa:
[...]

PR #121, opened 2026-08-19
Feat/110 compare images side by side

geourjoa:
https://github.com/archetype-pal/frontend/issues/110

Companion PRs https://github.com/archetype-pal/backend/pull/180


lanterno (changes requested):
Reviewed the full diff at `5776cf4a` and built the branch. The store, selection hook and tile-source fallback are well tested, bundle growth is confined to the `/compare` route chunks (all Mirador/MUI/emotion/redux chunks are dynamic-import-only; no page's client manifest references them), and the added deps are all Apache-2.0/MIT. Four things need fixing before merge.

**Blocking**

1. **The grid-view compare checkbox never fires.** `components/search/search-grid.tsx:590-593` does `onClick={(e) => { e.preventDefault(); e.stopPropagation(); }}`. Radix's `composeEventHandlers` skips its internal toggle when `event.defaultPrevented`, so `onCheckedChange` is never called. Verified with a vitest rendering `Checkbox` with that handler: 0 change events vs 1 without. Drop `preventDefault()` — the checkbox is a sibling of the `<Link>`, not inside it. Keep `stopPropagation` if you need it.

2. **Mirador is never unmounted.** `components/compare/mirador-viewer.tsx:30-43` discards the return value of `Mirador.viewer()`, but it is a `MiradorViewer` with an `unmount()` method (`mirador/src/lib/MiradorViewer.jsx`) — the comment saying there is no teardown API is wrong. Removing a chip re-runs the effect and calls `createRoot` on the same div again (React warns; the old root, Redux store, sagas and Mirador's internal OSD viewers stay alive), and leaving `/compare` never unmounts at all. Fix: `const v = Mirador.viewer(...)` and `return () => { cancelled = true; v?.unmount(); }`, plus `unmount(): void` in `types/mirador.d.ts`.

3. **Hydration mismatch on every page for anyone with staged items.** `stores/compare-store.ts:49-55` uses `persist` with sync `sessionStorage`, which hydrates at module import (verified: items present before first render). The server renders "Compare (0)" / "Add to Compare" in `header.tsx` and `compare-toggle-button.tsx`, the client's first render shows the real count → hydration error. `collection-context.tsx:59-68` deliberately loads in `useEffect` for this reason. Use `skipHydration: true` + `useCompareStore.persist.rehydrate()` in a mounted effect, or a `useHydrated` guard.

4. **The section toggle is cosmetic.** `proxy.ts` `SECTION_ROUTE_MAP` has no `compare: '/compare'`, and neither `CompareToggleButton`, the search checkboxes nor the page check `isSectionEnabled('compare')`. Disabling the section in the backoffice only hides the nav link. Add the route-map entry and gate the two entry points.

**Minor**

- `components/search/search-page.tsx:89-93` — `addItem` returns false for duplicates too, so selecting an already-staged manuscript shows "Compare is full".
- `lib/site-features.ts` says `ALL_SECTION_KEYS` must match the backend seed; backend #180 doesn't add `compare`, so on existing deployments `normalizeSectionOrder` appends it after Past Events until an admin re-saves.
- Mirador config sets no `language`; its UI stays English for fr/de users.
- The `manuscript-annotorious.tsx:474` behaviour widening and the `osd-iiif-tile-source.ts` scale-factor fallback are unrelated fixes that would be easier to review in their own PR. The fallback itself is SIPI-3.0-safe.

Tests: the 24 new/changed unit tests assert real behaviour, but nothing covers the UI wiring, which is where #1 and #2 live. A render test for the grid/table checkbox and a `MiradorViewer` mount/unmount test would catch both.

_Review assisted by Claude Code._

lanterno (approved):
Approving after pushing the fixes for the four blocking points myself, so this can land now. Three commits on top of `5776cf4a`:

- `0515497` **fix(compare): let the grid-view compare checkbox actually toggle** — `search-grid.tsx` no longer calls `preventDefault()` in the checkbox `onClick` (Radix skips its toggle when the event is default-prevented); `stopPropagation()` stays. `results-table.tsx` had no such handler. New `search-grid.test.tsx` cases: clicking the checkbox calls `toggle(id)`, and no checkbox renders without a selection.
- `c4072e3` **fix(compare): unmount Mirador on cleanup and pass the site locale** — `mirador-viewer.tsx` keeps the `Mirador.viewer()` instance and calls `unmount()` in the effect cleanup, so removing a chip or leaving `/compare` no longer leaks a React root, Redux store and OSD viewers; `types/mirador.d.ts` gains `unmount(): void` and `language?: string`; `compare/page.tsx` passes the active locale so the Mirador UI follows fr/de. New `mirador-viewer.test.tsx` (mocked `mirador`) covers boot config, unmount on component unmount, no reboot on an identical manifest array, and unmount-before-reboot order.
- `dc37c8d` **fix(compare): avoid hydration mismatch, gate on the section flag, clearer toasts** — `compare-store.ts` uses `skipHydration: true` and a new `compare-store-hydrator.tsx` rehydrates from `sessionStorage` in an effect (mounted in `app/layout.tsx`, same pattern as `collection-context`), so the server and first client render agree in the header and toggle button. `proxy.ts` `SECTION_ROUTE_MAP` gains `compare: '/compare'`; `CompareToggleButton` renders nothing and `search-page.tsx` hides the checkboxes and toolbar button when the `compare` section is disabled. Selecting an already-staged manuscript now says so (`search.compareAction.alreadyStaged`, en/fr/de) instead of "Compare is full". Tests: store ignores storage until `rehydrate()`; toggle button add/remove and hidden-when-disabled.

Verified: `pnpm lint` 0 errors, `prettier --check`, `tsc --noEmit`, `vitest run` 162 files / 1589 tests, and the PR's CI (lint, format, test, build, lighthouse) is green on `dc37c8d`. Mirador and its peers remain dynamic-import-only, so the bundle shape is unchanged from the budget this PR already refreshed.

**Left for a follow-up, not blocking**: the backend `DEFAULT_SITE_FEATURES` seed does not know the `compare` section yet, so on an existing deployment it appears last in the nav order until an admin re-saves the site features; and the `manuscript-annotorious.tsx` / `osd-iiif-tile-source.ts` changes bundled here are unrelated to compare and would have been easier to review separately.

_Fixes and review done with Claude Code._

PR #129, opened 2026-08-21
Bulk-edit graphs from Explore/Search

saad-mhmd:
Adds an Annotating-mode toggle, selection, and the existing per-image bulk-edit dialog to the Graphs search tab. Depends on the backend PR (feat-73/graph-search-id-lookup) for the `id__in` filter.


geourjoa:
@saad this is effectively a problem. Caching system are great but eventually we always came in a situation like that ^^

I think there is a way to do a partial update of the index of Meillisearch : 
https://www.meilisearch.com/docs/reference/api/documents/add-or-update-documents

I suggest two things to add/edit 
- Add a func that do an **incremental sync** on a dedicated document https://www.meilisearch.com/docs/reference/api/documents/add-or-update-documents
- On _post_save_ or _post_delete_ start a transaction, call as many times needed the incremental sync, close the transaction.

Maybe also add a cron to force sync daily or twice daily. 

@lanterno  what do you think on this dirty read problem ?

saad-mhmd:
Landed as archetype-pal/backend#190 (closes archetype-pal/backend#189) - event-driven incremental Meilisearch sync on Graph/GraphComponent saves.

saad-mhmd:
Commit \`e34a4df\` is a cherry-pick (\`git cherry-pick -x\`) of #139, which fixes a pre-existing bug on \`main\`: the Allograph/Hand \`SearchableSelect\` dropdown didn't respond to the mouse wheel when opened from inside a Sheet (only the scrollbar-thumb drag worked). It predates this PR — reproducible via the per-image annotation gallery on \`main\` — but this PR reuses the same component for the new bulk-edit sheet, so it's included here too rather than shipping a known bug. Same fix, not a duplicate change to review.

lanterno:
Reviewed, fixed and merging.

**Fixed in c6d2d78**
- **Graph lookup by id now refuses unrequested graphs.** Without the backend `id__in` filter the endpoint returns every graph, and the dialog would have opened and saved the whole table.
- **Components and positions are pruned only when a save changes the allograph.** Pruning on every save deleted legacy rows outside the schema on unrelated edits, like a hand change. The dialog is shared, so this also affected the per-image gallery.
- **Editorial graphs are left out of the edit for non-superusers,** with a toast saying how many. The write API hides them from those users, so their saves returned 404.
- **The strings this PR added are translated** in en/fr/de. Older English text in the dialog is unchanged.

**Earlier on this branch**
- Merged `main` in. c8eb590 ports the text-only results mode into the new graph card, because main's version lived only in the shared media card.

**Checks:** tsc, eslint and prettier are clean, and the full vitest suite passes (1555 tests). Backend #183, which this depends on, was conflict-resolved and merged first.

**Not changed here:** deleting a single editorial graph as a non-superuser still returns 404, because search documents don't carry the annotation type.

lanterno (approved):
Approved after the fixes in c6d2d78.

